PR #18, opened 2016-10-16
[5] User Log In

### LINKED ISSUE #5: User Log In
Should set up devise and allow admins to log in

esoterik:
Resolves #5.

Really barebones -- just a basic devise set up with a couple of feature specs. 

Let me know if anything else should be added here!


spec/features/users/log_in_spec.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.feature 'Authentication' do
```
orenyk: Should this file be called `authentication_spec.rb` instead to match the top-level `feature` param?

spec/features/users/log_in_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+  it 'enforces log in to access the app' do
+    visit '/'
+    expect(page).to have_content('sign in')
+  end
+  it 'allows users to log in' do # rubocop:disable RSpec/ExampleLength
```
orenyk: What is `RSpec/ExampleLength` set to? Would it be better to extract some of this to other methods so that this passes without disabling the cop?

esoterik:
Pushed an update, this should be good to go when the build passes.

I'll open up another issue for registration.

app/models/user.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+# User model class for all user types.
+class User < ApplicationRecord
```
orenyk: I had to Google this since I had no idea what `ApplicationRecord` was, although since the specs were passing I assumed it made sense. Guess I'd better read up on Rails 5 :stuck_out_tongue:.

spec/factories/users.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+FactoryGirl.define do
+  factory :user do
+    email { 'email@email.com' }
```
orenyk: Should we use a sequence here?

orenyk: As a follow up, we're not currently running any ActiveRecord validations so in principle it doesn't matter that this isn't a sequence right now, but we'll need to update the factory when we do add those in. Alternatively, authentication would be broken if we didn't have those validations (login with e-mail is difficult if multiple accounts have the same e-mail address) so maybe those do fall under this PR, and we should both test and implement those validations on the model level, as well as update the factory. What do you think?

spec/features/users/authentication_spec.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.feature 'Authentication' do
+  def log_in(user)
```
orenyk: A minor point, but I'd move this to the bottom of the spec file (and I'm not sure if it should even be included in the `RSpec.feature` block, but taking it out might also break the test). I think it's better if the specs are read first and then you scroll down to view any abstracted methods.

esoterik:
Okay, ready for rereview!

orenyk:
Based on our last convo, this still needs validations for e-mail uniqueness and model tests. One last thought - doesn't Devise include some validators somewhere? I'm not sure if it's just a regex but if it's included we don't need to test for it.

config/initializers/devise.rb
```diff
@@ -0,0 +1,293 @@
...
+# Use this hook to configure devise mailer, warden hooks and so forth.
+# Many of these configuration options can be set straight in your model.
+Devise.setup do |config|
+  # The secret key used by Devise. Devise uses this key to generate
+  # random tokens. Changing this key will render invalid all existing
+  # confirmation, reset password and unlock tokens in the database.
+  # Devise will use the `secret_key_base` as its `secret_key`
+  # by default. You can change it below and use your own secret key.
+  # config.secret_key = 'your_key_here'
+  config.secret_key = 'b0d7f10dab0977beed8c852a8da68f4d6de9045a1251216d915db23'\
```
orenyk: Wait, do we want this hardcoded? I feel like we want to just inherit from `secret_key_base` and make sure it's appropriately set when we run tests.

config/secrets.yml
```diff
@@ -1,5 +1,5 @@
 default: &default
-  secret_key_base: <%= ENV["SECRET_KEY_BASE"] %>
+  secret_key_base: '3c709746a2efc1e2e118cae6d6b86045f70eb1201478aa4166c9be7d830526c90e4cb5320b81085cc1321c3481f01c5fb344b45e3e7dd6d1f1e35a4a8475b0fb'
```
orenyk: This should only be in `development` and/or test, but since `production` and `staging` inherit from `default` this is not a good solution, I don't think.

orenyk:
:shipit:

PR #23, opened 2016-10-22
[7] Add facilities resources

### LINKED ISSUE #7: Add Facilities Resources
Buildings which have many suites which have many rooms should all be able to be created.

esoterik:
Resolves #7.

Only includes very basic attributes for all resources, simple validations on those attributes, and minimal creation forms.


app/controllers/application_controller.rb
```diff
@@ -1,4 +1,17 @@
 # frozen_string_literal: true
+# Base controller class
```
orenyk: Isn't this going to conflict with the changes in #18?

orenyk: As in, we've made changes to the same file in both branches, even if some of those changes are identical.

app/controllers/buildings_controller.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+# Controller for Buildings
+class BuildingsController < ApplicationController
+  before_action :set_building, only: :show
+
+  def set_building
```
orenyk: Shouldn't this be private?

app/controllers/rooms_controller.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+# Controller for Rooms
+class RoomsController < ApplicationController
+  before_action :set_room, only: :show
+
+  def set_room
```
orenyk: Same here (should be private, I think).

app/controllers/suites_controller.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+# Controller for Suites
+class SuitesController < ApplicationController
+  before_action :set_suite, only: :show
+
+  def set_suite
```
orenyk: Same here (private)

app/models/suite.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+# Model to represent the middle level of facility information
+# A suite is composed of mulitple rooms, which can be bedrooms or common rooms
```
orenyk: Typo (multiple), missing final period.

app/models/room.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+#
+# Model to represent the lowest-level of facility information
+#
+# Attributes:
+#   - number: The number of the building, a string
```
orenyk: Typo (`s/building/room`), missing final period?

app/models/building.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# Model to represent the highest-level of facility information
```
orenyk: These comments are all good (although mildly inconsistent in terms of punctuation :stuck_out_tongue:). Can we make sure that they fit into a standard Ruby documentation format, e.g. [RDoc](https://github.com/rdoc/rdoc) or [YARD](http://yardoc.org/)? If we use YARD we can actually host our documentation [here](http://www.rubydoc.info/) if we wanted (not necessary).

app/models/suite.rb
orenyk: Also, do we want to implement a room `type` yet or a new issue?

spec/features/buildings_spec.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.feature 'Building Creation' do
+  it 'succeeds' do
+    visit 'buildings/new'
+    fill_in 'building_name', with: 'Silliman'
+    expect { click_on 'Create' }.to change { Building.count }.from(0).to(1)
```
orenyk: One thing that's been recommended to me with acceptance tests is to also make assertions against the UI, not against the backend. For example, here we might check that a success flash is displayed and that the resulting page has an element indicating that the operation worked (e.g. a title if we go to the show view or a link if we go to the index view, etc). This keeps the acceptance tests as completely representative of the user experience, and you rely on your controller / views to do the right thing (and test them in isolation if necessary).

Thoughts?

spec/models/suite_spec.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.describe Suite, type: :model do
+  describe 'basic validations' do
+    it_behaves_like 'does not allow nil', :number
```
orenyk: Is there a reason why you decided to build your own validation checkers as opposed to using `shoulda_matchers`? I feel like they cover all of these test cases.

spec/support/shared_examples/validation_examples.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
```
orenyk: My other comment notwithstanding, these are really nicely done, you're getting pretty awesome at spec support stuff :smile:.

app/controllers/application_controller.rb
esoterik: Yup, I'll deal with the conflicts once #18 is merged

config/routes.rb
```diff
@@ -1,3 +1,6 @@
 # frozen_string_literal: true
 Rails.application.routes.draw do
+  resources :buildings
+  resources :suites
+  resources :rooms
```
esoterik: I think that we *might* want rooms outside of suites sometimes; what if someone wants to filter by largest bedrooms or something?

app/models/suite.rb
esoterik: I'll add the number of beds here (assuming that you meant common room vs bedroom by `type`)

config/routes.rb
esoterik: I'm not 100% confident that it would be a common use case, but I'm also not sold on having to deal with a nested resource. What would the other benefits of nesting be?

esoterik:
Things that still need to be done in this PR:
- [ ] `has_many :through` associations
- [ ] YARD documentation
- [ ] Determine if we're nesting rooms under suites
- [ ] `Room#type` and `Suite#size`

app/models/room.rb
```diff
@@ -0,0 +1,15 @@
...
+#     Required, must be unique within the suite the room belongs to
+#
+class Room < ApplicationRecord
+  belongs_to :suite
```
esoterik: `belongs_to through` associations aren't a thing -- it looks like we can use `delegate` to get similar behavior? (see this [stack overflow](http://stackoverflow.com/a/11457714)) I don't think it's an ideal solution, though...

esoterik: We could also just have a `belongs_to :building` and write a validation to ensure that the room's building matches the suite's building?

esoterik:
Okay, I think the only thing left is to decide if we're nesting rooms under suites!

Pushed an update with creator service objects and YARD documentation

orenyk:
Based on our last discussion, let's bump the nesting issue until a future date, and I think that given the simple nature of the CodeClimate issue, we can just accept it for now. Let's get the merge conflict resolved and then I'll do a final review. Thanks!

esoterik:
squashed and rebased!

spec/factories/buildings.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+FactoryGirl.define do
+  sequence :building_name do |n|
```
orenyk: As discussed, we're going to inline these sequences inside the actual `factory` block.

spec/features/rooms_spec.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.feature 'Room Creation' do
+  def fill_in_room_info(suite:, **attrs)
```
orenyk: Let's move this to the bottom.

orenyk: But it's also awesome :smile:

spec/models/room_spec.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.describe Room, type: :model do
+  describe 'basic validations' do
+    it { is_expected.to validate_presence_of(:number) }
+    it { is_expected.to validate_presence_of(:beds) }
+    it { is_expected.not_to allow_value(-1).for(:beds) }
+    it { is_expected.to belong_to(:suite) }
+    describe 'number uniqueness' do
```
orenyk: If it is easy and simple, I believe that `shoulda_matchers` offers a way to do this and we should switch.

spec/models/suite_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    it { is_expected.to validate_presence_of(:size) }
+    it { is_expected.not_to allow_value(-1).for(:size) }
+    it { is_expected.to allow_value(0).for(:size) }
+    it 'equals the number of beds in all rooms' do
+      size = 2
+      suite = FactoryGirl.create(:suite)
+      size.times { FactoryGirl.create(:room, beds: 1, suite: suite) }
+      expect(suite.size).to eq(size)
+    end
+    it 'increments on room addition' do
```
orenyk: As discussed, let's move these three tests to the Room model specs as they are and also add tests checking our callbacks using a mock Suite, e.g. make sure that calling `save` and `destroy` sends the appropriate message to the suite. We understand that this may be "overtesting" but we feel like it's an acceptable tradeoff.

app/views/buildings/new.html.erb
```diff
@@ -0,0 +1,5 @@
+<h1> New Building </h1>
+<%= simple_form_for @building do |f| %>
+  <%= f.input :name %>
```
esoterik: it doesn't look like it, we should open a separate issue

esoterik:
Fixed issues that were discussed in review, still need to add tests for the service objects.

esoterik:
pushed an update with specs for the creator service objects! probably needs one last read through before merging!

PR #42, opened 2016-11-21
[19] Add User roles

### LINKED ISSUE #19: Set up User Roles
We figure out what we want to use to manage permissions (cancancan or pundit).

Also, we should set up user roles as enums!

Blocked on #5.

esoterik:
Resolves #19
  - Add role enum to User class
  - Install pundit
  - Add policies for existing resources


esoterik:
No review necessary -- done in pairing

PR #47, opened 2016-11-22
[43] Add remaining CRUD actions to facilities resources

### LINKED ISSUE #43: Finish adding CRUD actions to housing inventory resources
We can currently only create and read, we need to be able to update and destroy.

orenyk:
Resolves #43

- Extract forms into partial templates
- Add `submit_button` helper method to appropriately label buttons
- Move some feature specs around

Open question:

1. What do we want to do after the `destroy` action? If we had an `index` action we'd often redirect there, but right now I just have it redirecting to the root path.
2. Do we actually want to use the modified resource when we return from a failed `update`? This came up since I was using the room number in the page title but if you try to set that number to an empty string, it also messes with the title when you re-render that action. Not sure what the best way forward is.


orenyk:
We will merge following a green build; review via pair programming.

PR #50, opened 2016-11-28
[36] CRUD for Draws

### LINKED ISSUE #36: Admins Can Create Draws
Admins should be able to add Suites and Students to a Draw. Blocked on #5, #7.

esoterik:
Resolves #36

### DIFF
--- a/app/models/draw.rb
+++ b/app/models/draw.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+#
+# Model to represent Housing Draws.
+#
+# @attr [String] name The name of the housing draw -- e.g. "Junior Draw 2016"
+# @attr [Array<User>] students The students in the draw.
+# @attr [Array<Suite>] suites The suites in the draw.
+class Draw < ApplicationRecord
+  has_many :students, class_name: 'User'
+  has_and_belongs_to_many :suites # rubocop:disable Rails/HasAndBelongsToMany
+
+  validates :name, presence: true
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -6,6 +6,7 @@ class User < ApplicationRecord
   devise :database_authenticatable, :registerable,
          :recoverable, :rememberable, :trackable, :validatable
 
+  belongs_to :draw
   validates :email, uniqueness: true
   validates :role, presence: true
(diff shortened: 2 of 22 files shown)


esoterik:
The builds are failing due to the has_and_belongs_to_many association -- the draw-suite association could also be a User has_many :suites, through: :draw? I'm unsure which will be better in the long run, because we will also likely need users to be associated to which ever Suite their Group selects.

esoterik:
This should probably also add all the deadlines necessary in Draws.

orenyk:
We can do that as a separate issue since I think it's going to involve a lot of separate testing.

spec/services/creators/draw_creator_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    params = instance_spy('ActionController::Parameters',
+                          to_h: { name: 'Silliman' })
+    expect(described_class.new(params).create![:object]).to \
+      be_instance_of(Draw)
+  end
+  it 'does not create when given invalid params' do
+    params = instance_spy('ActionController::Parameters', to_h: { name: nil })
+    expect(described_class.new(params).create![:object]).to be_nil
+  end
+  it 'returns a notice flash message' do
```
orenyk: Should we group this test and the one on L5 into a "success" context?

app/models/user.rb
```diff
@@ -6,6 +6,7 @@ class User < ApplicationRecord
   devise :database_authenticatable, :registerable,
          :recoverable, :rememberable, :trackable, :validatable
 
+  belongs_to :draw
```
esoterik: yeah, we should have a discussion about this before break, I think. I'll open up a new issue.

app/models/draw.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+#
+# Model to represent Housing Draws.
+#
+# @attr [String] name The name of the housing draw -- e.g. "Junior Draw".
+# @attr [Integer] year The year of the housing draw -- e.g. 2016.
```
orenyk: Lol I mean to add it in the name within the comment, not to add a separate attribute. Sorry for the confusion, let's revert to avoid unnecessary complexity.

PR #55, opened 2016-12-01
[37] Intent Declaration

### LINKED ISSUE #37: Students can submit Intent
Students should be able to opt out of the housing draw / application. Blocked on #5.

esoterik:
Resolves #37

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,7 +1,8 @@
 # frozen_string_literal: true
 # Users Controller class
 class UsersController < ApplicationController
-  before_action :set_user, only: %i(show edit update)
+  prepend_before_action :set_user, only: %i(show edit update edit_intent
+                                            update_intent)
 
   def show
   end
@@ -10,14 +11,16 @@ class UsersController < ApplicationController
   end
 
   def update
-    @user.assign_attributes(user_params)
-    if @user.save
-      flash[:notice] = 'User successfully updated.'
-      redirect_to @user
-    else
-      flash[:error] = 'Please review the errors below.'
-      render action: 'edit'
-    end
+    result = Updater.new(object: @user, name_method: :name,
+                         params: user_params).update
+    handle_action(action: 'edit', **result)
+  end
+
+  def edit_intent
+  end
+
+  def update_intent
+    update
   end
 
   private
@@ -36,6 +39,6 @@ class UsersController < ApplicationController
 
   def user_params
     params.require(:user).permit(:first_name, :preferred_name, :middle_name,
-                                 :last_name, :role, :email)
+                                 :last_name, :role, :email, :intent)
   end
 end
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -1,3 +1,4 @@
 <h1> User <%= @user.full_name %> </h1>
 <h2 class="user-role"> <%= @user.role %> </h2>
+<h2 class="user-intent"> <%= @user.intent %> </h2>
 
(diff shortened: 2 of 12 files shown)


app/controllers/users_controller.rb
```diff
@@ -1,7 +1,8 @@
 # frozen_string_literal: true
 # Users Controller class
 class UsersController < ApplicationController
-  before_action :set_user, only: %i(show edit update)
+  prepend_before_action :set_user, only: %i(show edit update edit_intent
```
orenyk: Don't we need to make this change in all of our other controllers as well? I'll open up a new issue.

config/routes.rb
```diff
@@ -6,4 +6,6 @@
   resources :suites
   resources :rooms
   resources :users
+  get 'users/:id/intent', to: 'users#edit_intent'
+  put 'users/:id/intent', to: 'users#update_intent'
```
orenyk: aw yea REST :smile:

db/migrate/20161201211657_add_intent_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddIntentToUsers < ActiveRecord::Migration[5.0]
+  def change
+    add_column :users, :intent, :integer
```
orenyk: I was going to comment about whether we wanted to set `null: false`, but then I realized it will make our reports of who submitted intent much easier. We just have to remember to reset intent every year as part of the setup process. I'll open an issue for that as well.

app/views/users/show.html.erb
```diff
@@ -1,3 +1,4 @@
 <h1> User <%= @user.full_name %> </h1>
 <h2 class="user-role"> <%= @user.role %> </h2>
+<h2 class="user-intent"> <%= @user.intent %> </h2>
```
orenyk: We may want to make this element contingent on the presence of an intent selection, otherwise it will still render even if there's no visible text.

spec/policies/user_policy_spec.rb
```diff
@@ -7,27 +7,30 @@
 
   context 'student' do
     let(:user) { FactoryGirl.build_stubbed(:user, role: 'student') }
-    permissions :show?, :update? do
+    permissions :show?, :update?, :edit?, :edit_intent?, :update_intent? do
```
orenyk: Lol, we forgot `edit`? Nice catch, were all of our other policy specs ok?

app/views/users/show.html.erb
esoterik: Actually, perhaps the better solution is adding a third "undeclared" value to the enum that it defaults to?

spec/policies/user_policy_spec.rb
esoterik: nope, we also forgot it for suites and possibly others. I'll open up a quick issue for it

app/views/users/show.html.erb
orenyk: Sure that's good.

spec/policies/user_policy_spec.rb
```diff
@@ -7,27 +7,30 @@
...
   end
 
   context 'housing rep' do
     let(:user) { FactoryGirl.build_stubbed(:user, role: 'rep') }
-    permissions :show?, :update? do
+    permissions :show?, :update?, :edit?, :edit_intent?, :update_intent? do
       it { is_expected.to permit(user, user) }
     end
-    permissions :show?, :create?, :destroy?, :update?, :index? do
+    permissions :show?, :create?, :destroy?, :update?, :index?,
```
orenyk: It just occurred to me, why are we checking `:index` for a specific record?

PR #72, opened 2016-12-09
[71] Add edit? Policy Tests

### LINKED ISSUE #71: Ensure permissions test coverage
In #55, we noticed that `edit?` was missing from the policy tests for User -- it's also definitely missing from suites. We need to make sure we have tests for all of our permissions!

esoterik:
Resolves #71


orenyk:
It does need to be rebased onto the latest master.

PR #73, opened 2016-12-17
[63] Fix order of resource before actions

### LINKED ISSUE #63: Update controller before actions
See #37 / #55, we need to use `prepend_before_action` for the `set_RESOURCE` before action so that it gets used by Pundit.

orenyk:
Resolves #63
- Change `before_action` to `prepend_before_action` in facilities controllers (building, suite, and room). This ensures that the specific resources are loaded into the relevant instance variables before Pundit checks for authorization to interact with those records


esoterik:
just needs as rebase!

orenyk:
Rebased! Will merge on CI passing.

PR #78, opened 2016-12-29
[49] Groups CRUD

### LINKED ISSUE #49: Students Can Form Groups
~~Blocked on #36.~~

esoterik:
Resolves #49. 

Still needs a bit of documentation but is otherwise ready for review!

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -22,7 +22,7 @@ class ApplicationController < ActionController::Base
   # @param [String] action The action to render when no object passed.
   #   (Creation / update failure, destruction success)
   # @param [String] path The path to redirect to when no object passed.
-  def handle_action(object:, msg:, action: nil, path: nil)
+  def handle_action(object:, msg:, action: nil, path: nil, **_)
     flash_type = msg.keys.first
     flash[flash_type] = msg[flash_type]
     redirect_to(object) && return if object
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+# Controller for Groups
+class GroupsController < ApplicationController
+  prepend_before_action :set_group, only: %i(show edit update destroy)
+  before_action :set_draw
+
+  def show; end
+
+  def new
+    @group = Group.new(draw: @draw)
+  end
+
+  def create
+    p = group_params.to_h
+    p[:leader_id] = current_user.id unless current_user.admin?
+    result = GroupCreator.new(p).create!
+    @group = result[:group] ? result[:group] : Group.new(draw: @draw)
+    handle_action(path: new_draw_group_path(@draw), **result)
+  end
+
+  def edit; end
+
+  def update
+    result = GroupUpdater.new(group: @group, params: group_params).update
+    handle_action(path: edit_draw_group_path(@draw, @group), **result)
+  end
+
+  def destroy
+    result = Destroyer.new(object: @group, name_method: :name).destroy
+    handle_action(**result)
+  end
+
+  private
+
+  def authorize!
+    if @group
+      authorize @group
+    else
+      authorize Group
+    end
+  end
+
+  def group_params
+    params.require(:group).permit(:size, :leader_id, member_ids: [])
+  end
+
+  def set_group
+    @group = Group.find(params[:id])
+  end
+
+  def set_draw
+    @draw = Draw.find(params[:draw_id])
+  end
+end
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+#
+# Model for Housing Groups
+#
+# @attr size [Integer] The room size that the Group wants.
+# @attr status [String] The state of the group (open, full, or locked)
+# @attr leader [User] The student that represents the Group.
+# @attr members [Array<User>] The members of the group, excluding the leader.
+# @attr draw [Draw] The Draw that this Group is in.
+class Group < ApplicationRecord
+  belongs_to :leader, class_name: 'User'
+  belongs_to :draw
+  has_many :members, class_name: 'User'
+
+  validates :size, presence: true, numericality: { only_integer: true,
+                                                   greater_than: 0 }
+  validates :status, presence: true
+  validates :leader, presence: true
+  validates :draw, presence: true
+
+  enum status: %w(open full locked)
+
+  after_save :add_leader_to_members
+
+  def name
+    "#{leader.name}'s Group"
+  end
+
+  private
+
+  def add_leader_to_members
+    members.append(leader) unless members.include? leader
+  end
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -7,6 +7,7 @@ class User < ApplicationRecord
          :recoverable, :rememberable, :trackable, :validatable
 
   belongs_to :draw
+  belongs_to :group
   validates :email, uniqueness: true
   validates :role, presence: true
   validates :first_name, presence: true
--- a/app/policies/group_policy.rb
+++ b/app/policies/group_policy.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+# Class for Group permissions
+class GroupPolicy < ApplicationPolicy
+  def show?
+    true
+  end
+
+  def index?
+    true
+  end
+
+  def create?
+    (user.draw && !user.group) || user.admin?
+  end
+
+  def edit?
+    record.leader == user || user.admin?
+  end
+
+  def destroy?
+    edit?
+  end
+
+  def update?
+    edit?
+  end
+
+  class Scope < Scope # rubocop:disable Style/Documentation
+    def resolve
+      scope
+    end
+  end
+end
--- a/app/services/creators/group_creator.rb
+++ b/app/services/creators/group_creator.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+#
+# Service object to create Groups.
+class GroupCreator < Creator
+  # Initialize a new GroupCreator
+  #
+  # @param params [#to_h] The params for the Group.
+  def initialize(params)
+    super(klass: Group, name_method: :name, params: params)
+    process_params
+  end
+
+  private
+
+  def process_params
+    @params = params.to_h.transform_keys(&:to_sym)
+    add_draw_to_params if params[:leader_id]
+    remove_blank_members if params[:member_ids]
+  end
+
+  def add_draw_to_params
+    @params.merge!(draw_id: User.find(params[:leader_id]).draw.id)
+  end
+
+  def remove_blank_members
+    @params[:member_ids] = params[:member_ids].reject(&:empty?)
+  end
+
+  def success(group)
+    { object: [group.draw, group], group: group,
+      msg: { notice: "#{group.name} created." } }
+  end
+end
--- a/app/services/updaters/group_updater.rb
+++ b/app/services/updaters/group_updater.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+#
+# Service object to update groups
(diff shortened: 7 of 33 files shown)


app/models/group.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+#
+# Model for Housing Groups
+#
+# @attr [Integer] size The room size that the Group wants.
+# @attr [String] status The state of the group (open, full, or locked)
+# @attr [User] leader The student that represents the Group.
+# @attr [Array<User>] members The members of the group, excluding the leader.
+# @attr [Draw] draw The Draw that this Group is in.
+class Group < ApplicationRecord
```
esoterik: this is missing the more complex validations (no more students than size, all students must be in the same draw, size must be available in the current draw, etc), but I think those should be done in a separate issue because this PR is already fairly long

spec/policies/group_policy_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+        it { is_expected.not_to permit(user, Group) }
+      end
+      context 'not in draw' do
+        it { is_expected.not_to permit(user, Group) }
+      end
+    end
+    permissions :index? do
+      it { is_expected.to permit(user, [group, other_group]) }
+    end
+    permissions :show? do
```
orenyk: Are we allowing users to view groups in other draws?

app/services/creators/group_creator.rb
```diff
@@ -0,0 +1,34 @@
...
+  def add_draw_to_params
+    @params.merge!(draw_id: User.find(params[:leader_id]).draw.id)
+  end
+
+  def remove_blank_members
+    @params[:member_ids] = params[:member_ids].reject(&:empty?)
+  end
+
+  def success(group)
+    { object: [group.draw, group], group: group,
```
orenyk: I was confused about this value for `object` but now I see you're using it to redirect to a nested resource, nice!

spec/services/creators/group_creator_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+  end
+
+  it 'does not create when given invalid params' do
+    params = instance_spy('ActionController::Parameters', to_h: {})
+    expect(described_class.new(params).create![:object]).to be_nil
+  end
+
+  def params_hash
+    draw = FactoryGirl.create(:draw_with_members, students_count: 1)
```
orenyk: Do we want to memoize these just in case this ever gets called twice? I know it's unlikely but seeing guaranteed resource creation within a helper method is a little troubling to me.

app/controllers/application_controller.rb
```diff
@@ -23,7 +23,7 @@ def index
   # @param [String] action The action to render when no object passed.
   #   (Creation / update failure, destruction success)
   # @param [String] path The path to redirect to when no object passed.
-  def handle_action(object:, msg:, action: nil, path: nil)
+  def handle_action(object:, msg:, action: nil, path: nil, **_)
```
orenyk: Why was this necessary?

esoterik: The method gets passed the splatted `results` from creator objects; this allows those results to have arbitrary keys that get used in the actual controller method (here, `:group` is the key that i used)

app/controllers/groups_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+  def set_group
+    @group = Group.find(params[:id])
+  end
+
+  def set_draw
+    @draw = Draw.find(params[:draw_id])
+  end
+
+  def set_user
+    @user = current_user
```
esoterik: this is actually totally unnecessary because `current_user` is available in views, will take it out

app/models/user.rb
```diff
@@ -7,6 +7,7 @@ class User < ApplicationRecord
          :recoverable, :rememberable, :trackable, :validatable
 
   belongs_to :draw
+  belongs_to :group
```
esoterik: hmm, I'm not sure. 

At the moment, leaders are not included in the members of a draw, so there's no conflict. Do you think it will help us in the long run to duplicate leaders in the members array? I'm still not sure that we would need the `has_one` because it would always be the same id as the `belongs_to`.

app/services/updaters/group_updater.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+# Service object to update groups
```
esoterik: I think we're somewhat inconsistent about this across the board; I think that the empty line is probably better so I'll add it

app/views/groups/_form.html.erb
```diff
@@ -0,0 +1,8 @@
+<%= simple_form_for [@draw, @group] do |f| %>
```
esoterik: not 100% sure, simple_form figures it out I think. (see this [stackoverflow](http://stackoverflow.com/questions/23945600/rails-4-action-routes-with-simple-form-and-shallow-nested-resources))

app/policies/group_policy.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+# Class for Group permissions
+class GroupPolicy < ApplicationPolicy
+  def show?
+    true
+  end
+
+  def index?
```
esoterik: yeah, I think that the index actions will end up being more complicated so they belong in separate issues.

db/migrate/20161224170703_create_groups.rb
```diff
@@ -0,0 +1,15 @@
+class CreateGroups < ActiveRecord::Migration[5.0]
+  def change
+    create_table :groups do |t|
+      t.integer :size, null: false, default: 1
+      t.integer :status, null: false, default: 0
+      t.belongs_to :leader, index: true, null: false
```
esoterik: Yeah, I was thinking that the leader would always be a student because they eventually get the permission to select a suite for the group.

spec/policies/group_policy_spec.rb
esoterik: I think it's useful for trying to predict what room sizes will be leftover for your draw

esoterik:
pushed an update!

app/models/user.rb
orenyk: I don't know if Rails will know how to assign the `group_id` for the `leader` unless they're included in the `members` array, since this `belongs_to` corresponds to the `has_many :members` in the Group model, whereas a `has_one :group` will correspond to the `belongs_to :leader` in the Group model. Now, if we include the leader as a member (probably through a callback or something), then that will provide the `leader.group` method and we shouldn't need to use `has_one :group` here, and if we did include it anyway I'm not sure if it would break anything.

Does that make sense?

app/views/groups/_form.html.erb
orenyk: Ohhh I didn't realize that it was a nested route when I first looked here, so that's why I was confused. Now it makes sense.

db/migrate/20161224170703_create_groups.rb
orenyk: Well, I'm guessing that in some situations the admins would be performing suite selection as well, but yea we're going to want them to assign a leader. We just have to make sure that it's easy to do.

spec/policies/group_policy_spec.rb
orenyk: Yea, but do they need to see which students are in each group or just the number of groups?

orenyk: I mean, if it's not sensitive info (who's living with who) then it doesn't matter, and we can always change it later anyway.

app/controllers/application_controller.rb
orenyk: Ohhh otherwise it was throwing some kind of mismatched parameter error? Got it.

app/models/group.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+#
+# Model for Housing Groups
+#
+# @attr [Integer] size The room size that the Group wants.
```
orenyk: So I think you've been doing YARD wrong 😕 - I'm pretty sure the format is `@attr attr_name [Type] description`, and until now the attribute name has been going _after_ the type. Let's fix all the new comments in this branch and I'll open an issue to update all our earlier documentation at some point.

spec/factories/groups.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+FactoryGirl.define do
+  factory :group do
+    size 1
+    status 'open'
```
orenyk: Shouldn't this actually be closed by default because a Group of size 1 is just the leader?

app/models/user.rb
esoterik: yup, will do!

spec/policies/group_policy_spec.rb
esoterik: yeah, I think we can change this later if necessary

app/models/group.rb
esoterik: ohno! I must have looked at an inaccurate reference when I started using it and never checked up on it

esoterik:
ok, should be good to go now!

app/controllers/groups_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+  def show; end
+
+  def new
+    @group = Group.new(draw: @draw)
+  end
+
+  def create
+    p = group_params.to_h
+    p[:leader_id] = current_user.id unless current_user.admin?
+    result = GroupCreator.new(p).create!
```
orenyk: I know this is probably a minor point, but aren't we passing in a `Hash` here instead of an `ActionController::Parameters` object as we've done in the other controllers / creators (and as we document in the `GroupCreator`)? With duck typing it shouldn't matter, but we should update the documentation in the `GroupCreator` to be accurate.

app/services/creators/group_creator.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+#
+# Service object to create Groups.
+class GroupCreator < Creator
+  # Initialize a new GroupCreator
+  #
+  # @param params [ActionController::Parameters] The params object from
```
orenyk: See my last comment.

spec/factories/groups.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+FactoryGirl.define do
+  factory :group do
+    size 1
+    status 'full'
+    association :leader, factory: :student_in_draw
+    draw { leader.draw }
```
orenyk: Do we need to add `members { [leader] }` as well, or do you want to do that when we add the validations / callbacks that ensure that the leader is also a member?

app/controllers/groups_controller.rb
esoterik: oooh, turns out YARD has a way to [denote](http://yardoc.org/types.html#see_results) duck types so I'll update it accordingly

spec/factories/groups.rb
esoterik: I added the callback already but not the validation, so I think we're okay

app/controllers/groups_controller.rb
orenyk: Excellent, nice find 😄

spec/factories/groups.rb
orenyk: Whoops I missed that, I'll go double check and then we should be good to go.

app/models/group.rb
```diff
@@ -0,0 +1,35 @@
...
+  after_create :add_leader_to_members
+
+  def name
+    "#{leader.name}'s Group"
+  end
+
+  private
+
+  def add_leader_to_members
+    members.append(leader)
```
orenyk: Hmm, do we want to maybe make this an `after_save` and ensure that the leader is _always_ a member? Also, if the leader was somehow already selected as a member (e.g. admin-created group) would this result in a doubled reference?

esoterik:
callback is now an `after_save` and checks if the leader is already a member before appending!

PR #82, opened 2016-12-30
[74] Fix pundit

### LINKED ISSUE #74: Figure out how Pundit works and if we're doing it wrong
Inspired by [this](https://github.com/YaleSTC/vesta/pull/55#discussion_r92873693), we're probably skipping authorization (and verification) for all `index` actions, and should also probably be doing something with the scope for index actions or at least something different from regular actions (maybe). Let's figure this out 😄.

esoterik:
Resolves #74


spec/policies/building_policy_spec.rb
```diff
@@ -13,11 +13,17 @@
     permissions :create?, :destroy?, :edit?, :update?, :index? do
```
orenyk: Should we be removing `index?` from here?

spec/policies/draw_policy_spec.rb
```diff
@@ -13,22 +13,31 @@
     permissions :create?, :destroy?, :update?, :index? do
```
orenyk: Same here

spec/policies/suite_policy_spec.rb
```diff
@@ -10,25 +10,34 @@
     permissions :show?, :index? do
```
orenyk: Same here

spec/policies/user_policy_spec.rb
```diff
@@ -10,28 +10,37 @@
     permissions :show?, :update?, :edit?, :edit_intent?, :update_intent? do
       it { is_expected.to permit(user, user) }
     end
-    permissions :show?, :create?, :destroy?, :update?, :index?,
+    permissions :show?, :create?, :destroy?, :update?,
                 :edit?, :edit_intent?, :update_intent? do
```
orenyk: Should we move `:edit?` up to the previous line for consistency? Also in a few other places.

orenyk:
We should also fix that CodeClimate issue.

esoterik:
We should use the scoping features when we actually get a collection that isn't just every record, I think, but considering we don't even have actual index actions right now I'm comfortable with authorizing against the class.

esoterik:
Also, I'm not sure why those code climate issues keep popping up -- they should also be getting caught by rubocop on our end, right?

esoterik:
ready for another review!

orenyk:
It's possible that our rubocop settings are different than theirs are? It is weird. I agree about the index / scoping thing, we can address that as necessary. Merging!

PR #87, opened 2017-01-03
[75] Prepare user model for IDR

### LINKED ISSUE #75: Clean up user model stuff for IDR
We no longer need a preferred and middle name as that will be handled during import / update from IDR. Also remove the ability to edit / update all attributes except for gender, which we will allow students to override locally.

esoterik:
Resolves #75
  - Removes middle names and preferred names from users
  - Adds gender enum to users

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -13,15 +13,17 @@ class User < ApplicationRecord
   validates :first_name, presence: true
   validates :last_name, presence: true
   validates :intent, presence: true
+  validates :gender, presence: true
 
   enum role: %w(student admin rep)
   enum intent: %w(undeclared on_campus off_campus)
+  enum gender: %w(non-binary female male)
 
-  # Returns the user's preferred first name
+  # Returns the user's preferred name
   #
-  # @return [String] Preferred name if set, otherwise first name.
+  # @return [String] Preferred name
   def name
-    preferred_name || first_name
+    first_name
   end
 
   # Returns the user's preferred full name
@@ -29,16 +31,5 @@ class User < ApplicationRecord
   # @return [String] Preferred name if set, otherwise first name, plus last name
   def full_name
     "#{name} #{last_name}"
-  end
-
-  # Returns the user's full legal name
-  #
-  # @return [String] The first, middle (if present), and last name of the user.
-  def legal_name
-    if middle_name
-      "#{first_name} #{middle_name} #{last_name}"
-    else
-      "#{first_name} #{last_name}"
-    end
   end
 end
--- a/app/views/users/edit.html.erb
+++ b/app/views/users/edit.html.erb
@@ -1,10 +1,6 @@
 <h1> Editing <%= @user.name %> </h1>
 <%= simple_form_for @user do |f| %>
-  <%= f.input :email %>
-  <%= f.input :first_name %>
-  <%= f.input :preferred_name %>
-  <%= f.input :middle_name %>
-  <%= f.input :last_name %>
-  <%= f.select :role, collection: User.roles.keys %>
+  <%= f.select :gender, collection: User.genders.keys %>
+  <%= f.select :role, collection: User.roles.keys if current_user.admin? %>
   <%= f.submit 'Save' %>
 <% end %>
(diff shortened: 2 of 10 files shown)


app/models/user.rb
```diff
@@ -12,15 +12,17 @@ class User < ApplicationRecord
...
   enum role: %w(student admin rep)
   enum intent: %w(undeclared on_campus off_campus)
+  enum gender: %w(non-binary female male)
 
-  # Returns the user's preferred first name
+  # Returns the user's preferred name
   #
-  # @return [String] Preferred name if set, otherwise first name.
+  # @return [String] Preferred name
   def name
```
orenyk: Woot! I like how we don't have to change things _everywhere_ for this 😄

app/models/user.rb
```diff
@@ -12,15 +12,17 @@ class User < ApplicationRecord
   validates :first_name, presence: true
   validates :last_name, presence: true
   validates :intent, presence: true
+  validates :gender, presence: true
 
   enum role: %w(student admin rep)
   enum intent: %w(undeclared on_campus off_campus)
+  enum gender: %w(non-binary female male)
```
orenyk: According to the data dictionary, the values we're getting from IDR are:

> Valid values are NULL (unknown), "N" (Not Declared), "F" (Female) or "M" (Male)

How should we handle "N" and NULL cases? "N" doesn't necessarily feel like non-binary to me, but you probably have a better idea about this than I do.

spec/factories/users.rb
```diff
@@ -5,6 +5,7 @@
     password { 'passw0rd' }
     first_name { 'First' }
     last_name { 'Last' }
+    gender { 0 }
```
orenyk: Why are you using the enum value and not the string?

app/views/users/edit.html.erb
```diff
@@ -1,10 +1,6 @@
 <h1> Editing <%= @user.name %> </h1>
 <%= simple_form_for @user do |f| %>
-  <%= f.input :email %>
-  <%= f.input :first_name %>
-  <%= f.input :preferred_name %>
-  <%= f.input :middle_name %>
-  <%= f.input :last_name %>
-  <%= f.select :role, collection: User.roles.keys %>
+  <%= f.select :gender, collection: User.genders.keys %>
```
orenyk: Not that we need to change this now, but we're going to need to add some labels to these forms eventually 😛.

spec/features/users/edit_spec.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.feature 'User Editing' do
+  let(:user) { FactoryGirl.create(:student) }
```
orenyk: Should we specify the gender to make it clear that it's not already `male`?

orenyk:
Weird, one of my comments is claiming it's on an outdated diff but the change is still there (it's on the factory).

app/models/user.rb
esoterik: I think defaulting to non-binary is fine

esoterik:
updated the feature test!

spec/factories/users.rb
```diff
@@ -5,6 +5,7 @@
     password { 'passw0rd' }
     sequence(:first_name) { |n| "First_#{n}" }
     last_name { 'Last' }
+    gender { 0 }
```
orenyk: This is the line that it claimed was outdated; is there a reason why we're using the integer value and not the string here?

esoterik: nope, switched it to the string

orenyk:
Actually, it looks like that broke some things on Travis...

spec/factories/users.rb
```diff
@@ -5,6 +5,7 @@
     password { 'passw0rd' }
     sequence(:first_name) { |n| "First_#{n}" }
     last_name { 'Last' }
+    gender { 'non_binary' }
```
orenyk: Oh this is supposed to be `'non-binary'` (dash, not underscore). Feel free to change the Enum definition instead.

esoterik:
sorry about that, should be good to go now

Issue #56, opened 2016-12-02
Students Can Request to Join Groups


esoterik:
Blocked on #49

esoterik:
blocking on #81 due to the introduction of Memberships

esoterik:
implementation details:
- Memberships are getting a `status` enum (approved, pending, and invited)
- `group.memberships_count` will become a conditional counter cache that excludes `pending` memberships

esoterik:
okay, running into problems/pain:

The conditional counter cache is fine, BUT with the new statuses on memberships, all auto-created memberships have status `requested` (e.g. memberships created via `group.members << student`). This means that we lose a lot of the syntactic sugar of `has_many :through` 

I think there's other problems, too, but I think a solution to this might be adding separate associations on `Group` for pending memberships. This way minimizes additional callbacks and let's us keep the default membership status as something unconfirmed -- another solution is to just make the default status the accepted/confirmed one, but that feels unsafe to me.

@orenyk, thoughts?

orenyk:
Per phone discussion, we are good to set "confirmed/accepted" as the default as we'll be explicitly creating memberships in all the other situations.

Issue #51, opened 2016-11-29
Intent Deadline


esoterik:
Draws should have a deadline for intent submission. After this deadline passes, students in the draw who have not submitted intent will default to living on campus (or, potentially make this configurable by admins)

orenyk:
After reviewing some of our notes from the Residential College admins, it looks like at least one is opposed to any kind of intent assumption (Trumbull) and would just have the admin bother the students offline until they either indicated intent themselves or verbally told the admin who would then make the selection on behalf of the student.

We can still add this deadline and ultimately use it for sending reminder e-mails, but we can also expand this issue to include the admin report of students/intent as well as their ability to indicate intent. Does that make sense?

orenyk:
Ok, thinking some more about this - would we prefer to keep the admin reports as separate routes from the student ones or just have a single route with different content depending on `current_user`? It gets even more confusing because it may also depend on the stage of the draw process we're in.

My first thought was just to use a nested route, `/draws/:id/students` and the columns / actions would change based on who was viewing and what was going on (e.g. pre-intent, we'd see the student name, their intent (maybe as an inline form with a dropdown to update the value as an admin), during the group phase we could see the students and their group (again, potentially as a dropdown), and maybe the same during the draw itself depending on how we handle oversubscription). At the same token, I figured we could use that same route for students to see everyone else's status, although now that I'm thinking about it, I'm not sure why that route would be meaningful for students since we'll likely give them something like `/draws/:id/groups` for group selection.

I'm also considering a general dashboard / status page for admins, but that would be a separate issue. Any thoughts on how you see this playing out would be great, thanks!

orenyk:
Some notes from a phone conversation:

1. General dashboard with metrics and links through to specific reports relating to each aspect (e.g. intent, groups, etc) -->  we can also include buttons to send reminder e-mails manually.
2. No defaulting any kind of intent selection

Integration test:

1. visit draw page
2. test to see that intent metrics are correct
3. click through to intent report (`/draws/:id/intent`)
4. test for specific content
5. test that admins can set intent via inline forms

orenyk:
Ok, basic status page integration spec set up with intent metrics, although the query used is super janky and needs to be extracted into a query object. I'll do that and work on the reports next.

orenyk:
[...]

PR #89, opened 2017-01-04
[81] Group Validations

### LINKED ISSUE #81: Add Group Validations
- Draw that the group belongs to should match the leader and all members
- The size must match an available room size in the Draw (this might be complicated for situations where suites in multiple draws are all taken during an earlier draw)
- Number of members + leader must not exceed the size
- Open/full status should be accurate
- Groups can only be locked if they are full

~~Blocked on #49~~

esoterik:
Somewhat of a WIP

Resolves #81

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -10,25 +10,65 @@
 class Group < ApplicationRecord
   belongs_to :leader, class_name: 'User'
   belongs_to :draw
-  has_many :members, class_name: 'User'
-
-  validates :size, presence: true, numericality: { only_integer: true,
-                                                   greater_than: 0 }
-  validates :status, presence: true
-  validates :leader, presence: true
-  validates :draw, presence: true
+  has_many :memberships
+  has_many :members, through: :memberships, source: :user
 
   enum status: %w(open full locked)
 
-  after_save :add_leader_to_members
+  validates :draw, presence: true
+  validates :status, presence: true
+  validates :size, presence: true,
+                   inclusion: { in: ->(g) { g.draw.suite_sizes },
+                                if: ->(g) { g.draw.present? } }
+  validates :leader, presence: true, inclusion: { in: ->(g) { g.members } }
+
+  validate :validate_members_count, if: ->(g) { g.size.present? }
+  validate :validate_status, if: ->(g) { g.size.present? }
+
+  before_validation :add_leader_to_members, if: ->(g) { g.leader.present? }
 
   def name
     "#{leader.name}'s Group"
   end
 
+  # Updates the status to match the group size (open when fewer members than
+  # the size, and full when they are the same)
+  def update_status
+    if memberships_count < size
+      update(status: 'open')
+    elsif memberships_count == size
+      update(status: 'full')
+    end
+  end
+
   private
 
   def add_leader_to_members
-    members.append(leader) unless members.include? leader
+    members << leader unless members.include? leader
+  end
+
+  def validate_members_count
+    return unless memberships_count > size
+    errors.add :members, "can't be greater than the size (#{size})"
+  end
+
+  def validate_status
+    return unless status_changed?
+    case status
+    when 'open'
+      validate_open
+    else
+      validate_full_locked
+    end
+  end
+
+  def validate_open
+    return unless memberships_count >= size
+    errors.add :status, 'can only be open when fewer members than size'
+  end
+
+  def validate_full_locked
+    return unless memberships_count != size
+    errors.add :status, "can only be #{status} when members equal size"
   end
 end
--- a/app/models/membership.rb
+++ b/app/models/membership.rb
@@ -1,0 +1,56 @@
+# frozen_string_literal: true
+#
+# Model for relationships between Groups and Users.
+# Updates the group status on creation and destruction.
+# Validations:
+#   - Users can only join groups in their draw
+#   - Groups can only be joined when they are open
+#   - Users must have declared on_campus intent
+#
+# @attr group [Group] The group of the membership.
+# @attr user [User] The user of the membership.
+class Membership < ApplicationRecord
+  belongs_to :group, counter_cache: true
+  belongs_to :user
+
+  validates :group, presence: true
+  validates :user, presence: true
+  validate :matching_draw, if: ->(m) { m.user.present? && m.group.present? }
+  validate :user_on_campus, if: ->(m) { m.user.present? }
+  validate :group_is_open, if: ->(m) { m.group.present? }, on: :create
+
+  before_destroy ->(m) { throw(:abort) if m.readonly? }
+  before_update ->(m) { throw(:abort) if m.readonly? }
+  before_update :freeze_group_and_user
+  after_create :update_group_status
+  after_destroy :update_group_status
+
+  def readonly?
+    group.locked?
+  end
+
+  private
+
+  def update_group_status
+    group.update_status
+  end
+
+  def freeze_group_and_user
+    return unless group_id_changed? || user_id_changed?
+    throw(:abort)
+  end
+
+  def matching_draw
+    return if user.draw == group.draw
+    errors.add :user, 'Member draw and group draw must match'
+  end
+
+  def group_is_open
+    errors.add :group, 'Group must be open' unless group.open?
+  end
+
+  def user_on_campus
+    return if user.on_campus?
+    errors.add :user, 'User must be living on campus to join a group'
+  end
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -7,7 +7,9 @@ class User < ApplicationRecord
          :recoverable, :rememberable, :trackable, :validatable
 
   belongs_to :draw
-  belongs_to :group
+  has_one :membership
+  has_one :group, through: :membership
+
   validates :email, uniqueness: true
   validates :role, presence: true
   validates :first_name, presence: true
--- a/spec/factories/groups.rb
+++ b/spec/factories/groups.rb
@@ -2,8 +2,25 @@
 FactoryGirl.define do
   factory :group do
     size 1
-    status 'full'
-    association :leader, factory: :student_in_draw
+    association :leader, factory: :student_in_draw, intent: 'on_campus'
     draw { leader.draw }
+    members { [leader] }
+
+    factory :full_group do
+      size 2
+      after(:build) do |g|
+        g.draw.suites << create(:suite_with_rooms, rooms_count: g.size)
+        (g.size - g.members.size).times do
+          g.members << create(:student, draw: g.draw)
+        end
+      end
+    end
+
+    factory :open_group do
+      size 2
+      after(:build) do |g|
+        g.draw.suites << create(:suite_with_rooms, rooms_count: g.size)
+      end
+    end
   end
 end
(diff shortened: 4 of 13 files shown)


esoterik:
working on this has called to attention that its annoying to build test data (to test memberships you need a draw, a suite of appropriate size in that draw, users in that draw, and a group in that draw); I'm not sure what the best solution to this will be in the end. 

Most of these tests end up persisting data and I think I'd like to refactor them away from that, but that's difficult to do without building brittle tests in most of these cases.

@orenyk let me know what you think! (a full review is probably overkill at this point considering the refactoring that I think will likely be necessary)

spec/models/group_spec.rb
```diff
@@ -13,15 +13,42 @@
...
-      expect(group.name).to include(leader.name)
+  describe 'size validations' do
+    it 'must be available in the draw' do
+      group = FactoryGirl.build(:group)
+      allow(group.draw).to receive(:suite_sizes).and_return([1])
+      group.size = 2
+      expect(group.valid?).to be_falsey
+    end
+    it 'must not have more members than the size' do
+      group = FactoryGirl.create(:group, size: 1)
```
orenyk: Oh is it because the `member_id`'s are no longer record attributes but come from the `memberships` table?

spec/models/membership_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    it { is_expected.to validate_presence_of(:user) }
+    it { is_expected.to validate_presence_of(:group) }
+  end
+
+  describe 'group draw and user draw must match' do
+    it do
+      user = FactoryGirl.create(:student)
+      group = FactoryGirl.create(:open_group)
+      membership = FactoryGirl.build(:membership, user: user, group: group)
+      expect(membership.valid?).to be_falsey
```
orenyk: Is this because the user has no draw? Would it be better to specify a different draw? I feel like we don't necessarily need to persist here, just stub out the `draw` method for both the `group` and `user` objects.

spec/models/group_spec.rb
```diff
@@ -13,15 +13,42 @@
...
+
+  describe 'status validations' do
+    it 'can only be locked if the number of members match the size' do
+      group = FactoryGirl.build(:group, size: 2)
+      group.draw.suites << FactoryGirl.create(:suite_with_rooms, rooms_count: 2)
+      group.status = 'locked'
+      expect(group.valid?).to be_falsey
+    end
+    it 'cannot be full when there are less members than the size' do
+      group = FactoryGirl.build(:group, size: 2)
```
orenyk: Any reason why you didn't use the `:open_group` factory here? You can use `group.size` to set the `rooms_count` parameter.

spec/models/group_spec.rb
```diff
@@ -13,15 +13,42 @@
...
+      group = FactoryGirl.create(:group, size: 1)
+      user = FactoryGirl.create(:student, draw: group.draw)
+      expect { group.members << user }.to \
+        raise_exception(ActiveRecord::RecordInvalid)
+    end
+  end
+
+  describe 'status validations' do
+    it 'can only be locked if the number of members match the size' do
+      group = FactoryGirl.build(:group, size: 2)
```
orenyk: Same here, you could use the `open_group` factory instead. That would also take care of the suite creation for you.

app/models/group.rb
```diff
@@ -10,25 +10,64 @@
...
+  validate :validate_status, if: proc { |g| g.size.present? }
 
-  after_save :add_leader_to_members
+  before_validation :add_leader_to_members, if: proc { |g| g.leader.present? }
 
   def name
     "#{leader.name}'s Group"
   end
 
+  def update_status
```
esoterik: Right now either memberships exist or they don't; editing won't be an option right now. I think this will change in #56.

app/models/membership.rb
```diff
@@ -0,0 +1,38 @@
...
+  belongs_to :group, counter_cache: true
+  belongs_to :user
+
+  validates :group, presence: true
+  validates :user, presence: true
+  validate :matching_draw, if: proc { |m| m.user.present? && m.group.present? }
+
+  validate :group_is_open, if: proc { |m| m.group.present? }, on: :create
+
+  after_create :update_group_status
```
esoterik: I think we should depend on the group status being accurate. A single membership shouldn't need to know how many other members there are in the group. 

Also, I'm not sure how things would get messed up on modification; right now there isn't anything to modify. I think we can move this conversation into how we handle #56, but I don't think it will be too much of a problem.

I added the count vs size validation back into groups, though.

app/models/user.rb
```diff
@@ -7,7 +7,9 @@ class User < ApplicationRecord
          :recoverable, :rememberable, :trackable, :validatable
 
   belongs_to :draw
-  belongs_to :group
+  has_one :membership
```
esoterik: Yeah I think we deal with it later; shouldn't be too bad to change this to a `has_many :through` eventually

spec/factories/groups.rb
```diff
@@ -2,8 +2,25 @@
...
     association :leader, factory: :student_in_draw
     draw { leader.draw }
+    members { [leader] }
+
+    factory :full_group do
+      size 2
+      after(:build) do |g|
+        g.draw.suites << create(:suite_with_rooms, rooms_count: g.size)
+        (g.size - g.members.size).times do
+          g.members << create(:student, draw: g.draw)
```
esoterik: yup!

spec/factories/groups.rb
```diff
@@ -2,8 +2,25 @@
...
+        end
+      end
+    end
+
+    factory :open_group do
```
esoterik: yeah I think we leave this for now; I guess we could throw an exception or something but it seems like overkill

spec/models/group_spec.rb
esoterik: both, I think

spec/models/membership_spec.rb
esoterik: if we don't persist, `membership.valid?` fails because the user and the group don't exist

esoterik:
okay, ready for another review!

esoterik:
@orenyk

spec/models/membership_spec.rb
orenyk: fair enough

spec/factories/groups.rb
orenyk: yea agreed, just something that popped out at me

app/models/user.rb
orenyk: 👍

app/models/membership.rb
orenyk: Yes agreed, this was a brain fart on my part

app/models/group.rb
orenyk: Right, but should we maybe prevent the modification of a membership's group or user following creation? I feel like we could do that with a simple validation to make sure that neither of those associations are ever modified except during create. What do you think?

app/models/group.rb
```diff
@@ -10,25 +10,65 @@
...
+                   inclusion: { in: ->(group) { group.draw.suite_sizes },
+                                if: ->(group) { group.draw.present? } }
   validates :status, presence: true
-  validates :leader, presence: true
+  validates :leader, presence: true,
+                     inclusion: { in: ->(group) { group.members } }
   validates :draw, presence: true
 
-  enum status: %w(open full locked)
+  validate :validate_members_count, if: proc { |g| g.size.present? }
```
orenyk: Any reason to use the `proc` notation here?

app/models/membership.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+#
+# Model for relationships between Groups and Users.
+# Updates the group status on creation and destruction.
+# Validations:
+#   - Users can only join groups in their draw
+#   - Groups can only be joined when they are open
```
orenyk: We should document that we're validating against intent as well.

app/models/membership.rb
```diff
@@ -0,0 +1,48 @@
...
+  validates :user, presence: true
+  validate :matching_draw, if: ->(m) { m.user.present? && m.group.present? }
+  validate :user_on_campus, if: ->(m) { m.user.present? }
+  validate :group_is_open, if: ->(m) { m.group.present? }, on: :create
+
+  before_destroy ->(m) { throw(:abort) if m.readonly? }
+  after_create :update_group_status
+  after_destroy :update_group_status
+
+  def readonly?
```
orenyk: Oh this is excellent

spec/models/group_spec.rb
```diff
@@ -13,15 +13,40 @@
...
-      allow(group).to receive(:leader).and_return(leader)
-      expect(group.name).to include(leader.name)
+  describe 'size validations' do
+    it 'must be available in the draw' do
+      group = FactoryGirl.build(:group)
+      allow(group.draw).to receive(:suite_sizes).and_return([1])
+      group.size = 2
+      expect(group.valid?).to be_falsey
+    end
+    it 'must not have more members than the size' do
```
orenyk: Hm, is this testing the group validations or the membership validations? I feel like we'd fail on the membership size since the group is full, so we'd never create the membership, therefore not update the counter cache, and therefore not hit the member count validation on the group. This is opening up a can of worms in my head - we can test this by creating an invalid group whose status is incorrect (thereby allowing the membership to be created), but then we end up with a borked db which sounds like a terrible idea.

I guess the membership count vs size validation is useful in ensuring that you can't reduce the size of a group below its memberships, so maybe that's what we should actually be testing? What do you think?

esoterik: that sounds like a better test!

spec/models/membership_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+    end
+  end
+
+  describe 'cannot change group' do
+    it do
+      group = FactoryGirl.create(:full_group)
+      membership = group.memberships.last
+      new_leader = FactoryGirl.create(:student, draw: group.draw)
+      membership.group = FactoryGirl.create(:open_group, leader: new_leader)
```
esoterik: this highlights some ugliness from how our `Group` factory is set up: we can't override the draw by itself because we end up with a mismatched draw between the group and the leader.

I tried to fix it but couldn't get anything working; passing optional attributes to association factories isn't easy.

app/models/group.rb
```diff
@@ -10,25 +10,65 @@
...
 
   def name
     "#{leader.name}'s Group"
   end
 
+  # Updates the status to match the group size (open when fewer members than
+  # the size, and full when they are the same)
+  def update_status
+    if memberships_count < size
+      update_attributes(status: 'open')
```
orenyk: While working on #51 I discovered that `update` is an alias for `update_attributes`, so since it's both shorter and cleaner why don't we use that and continue moving forward?

spec/factories/groups.rb
```diff
@@ -2,8 +2,25 @@
 FactoryGirl.define do
   factory :group do
     size 1
-    status 'full'
-    association :leader, factory: :student_in_draw
+    association :leader, factory: :student_in_draw, intent: 'on_campus'
     draw { leader.draw }
```
orenyk: I just noticed (and this touches on our discussion earlier) - so if we override the leader with a student outside of any draw, then our factory will create an invalid group? Not a problem, per se, just an observation.

spec/models/group_spec.rb
```diff
@@ -13,15 +13,40 @@
...
+      group = FactoryGirl.create(:full_group)
+      group.draw.suites << FactoryGirl.create(:suite_with_rooms, rooms_count: 1)
```
orenyk: Can't we just stub `groups.draw.suite_sizes` as in the example above rather than persisting a new suite? Either way we should be consistent.

esoterik:
should be good to go!

PR #95, opened 2017-01-06
[51] Add status enum to Draws with intent deadline

### LINKED ISSUE #51: Intent Deadline

orenyk:
Resolves #51
- Add status enum to draws (current values 'draft' and 'pre-lottery', with 'lottery' and 'post-lottery' to be implemented in the future)
- Add button to initiate draw on draw show view with appropriate route
- Add DrawStarter to check validity of draw setup and handle transition to pre-lottery stage.

### DIFF
--- a/app/controllers/draws_controller.rb
+++ b/app/controllers/draws_controller.rb
@@ -1,11 +1,12 @@
 # frozen_string_literal: true
+#
 # Controller for Draws
 class DrawsController < ApplicationController
-  before_action :set_draw, only: [:show, :edit, :update, :destroy]
+  prepend_before_action :set_draw, only: [:show, :edit, :update, :destroy,
+                                          :activate]
+  before_action :calculate_metrics, only: [:show, :activate]
 
-  def show
-    @intent_metrics = IntentMetricsQuery.call(@draw)
-  end
+  def show; end
 
   def new
     @draw = Draw.new
@@ -31,6 +32,11 @@ class DrawsController < ApplicationController
     handle_action(**result)
   end
 
+  def activate
+    result = DrawActivator.activate(draw: @draw)
+    handle_action(action: 'show', **result)
+  end
+
   private
 
   def authorize!
@@ -42,10 +48,15 @@ class DrawsController < ApplicationController
   end
 
   def draw_params
-    params.require(:draw).permit(:name, suite_ids: [], student_ids: [])
+    params.require(:draw).permit(:name, :intent_deadline, suite_ids: [],
+                                                          student_ids: [])
   end
 
   def set_draw
     @draw = Draw.find(params[:id])
+  end
+
+  def calculate_metrics
+    @intent_metrics = IntentMetricsQuery.call(@draw)
   end
 end
--- a/app/mailers/student_mailer.rb
+++ b/app/mailers/student_mailer.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+#
+# Mailer class for user e-mails
+class StudentMailer < ApplicationMailer
+  default from: 'no-reply@vesta.app'
+
+  # Send initial invitation to students in a draw
+  #
+  # @attr user [User] the user to send the invitation to
+  def draw_invitation(user)
+    @user = user
+    @intent_deadline = user.draw.intent_deadline
+    @res_college = OpenStruct.new(name: 'College', dean: 'Dean Vesta',
+                                  vesta_url: 'https://vesta.app/',
+                                  admin_email: 'admin@vesta.app')
+    mail(to: @user.email, subject: 'The housing process has begun')
+  end
+end
--- a/app/models/draw.rb
+++ b/app/models/draw.rb
@@ -2,16 +2,47 @@
 #
 # Model to represent Housing Draws.
 #
-# @attr [String] name The name of the housing draw -- e.g. "Junior Draw 2016"
-# @attr [Array<User>] students The students in the draw.
-# @attr [Array<Suite>] suites The suites in the draw.
+# @attr name [String] The name of the housing draw -- e.g. "Junior Draw 2016"
+# @attr students [Array<User>] The students in the draw.
+# @attr suites [Array<Suite>] The suites in the draw.
+# @attr status [String] The status / phase of the draw (draft, pre_lottery,
+#   TODO: lottery, post_lottery). Note the use of underscores in the status
+#   strings; this prevents some unpleasantness with the helper methods.
 class Draw < ApplicationRecord
   has_many :students, class_name: 'User'
   has_and_belongs_to_many :suites # rubocop:disable Rails/HasAndBelongsToMany
 
   validates :name, presence: true
+  validates :status, presence: true
+
+  enum status: %w(draft pre_lottery)
 
   def suite_sizes
     suites.map(&:size).uniq
+  end
+
+  # Query method to see if a draw has at least one student.
+  #
+  # @return [Boolean] whether or not the draw has at least one student
+  def students?
+    student_count.positive?
+  end
+
+  # Query method to see if a draw has enough beds for its students.
+  #
+  # @return [Boolean] whether or not the draw has as many or more beds than it
+  #   has students
+  def enough_beds?
+    bed_count >= student_count
+  end
+
+  private
+
+  def student_count
+    @student_count ||= students.count
+  end
+
+  def bed_count
+    @bed_count ||= suites.sum(:size)
   end
 end
--- a/app/policies/draw_policy.rb
+++ b/app/policies/draw_policy.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+#
 # Class for Draw permissions
 class DrawPolicy < ApplicationPolicy
   def show?
@@ -7,6 +8,10 @@ class DrawPolicy < ApplicationPolicy
 
   def update?
     user.rep? || user.admin?
+  end
+
+  def activate?
+    user.admin? && record.draft?
   end
 
   class Scope < Scope # rubocop:disable Style/Documentation
--- a/app/services/draw_activator.rb
+++ b/app/services/draw_activator.rb
@@ -1,0 +1,78 @@
+# frozen_string_literal: true
+#
+# Service object to handle the activation of a draw. Checks to make sure the
+# draw has the correct status, updates the status, and sends e-mail invitations
+# to students in the draw.
+class DrawActivator
+  # Class method to permit calling :activate on the class without instantiating
+  # the service object directly
+  #
+  # @attr draw [Draw] the draw in question
+  def self.activate(draw:)
+    new(draw: draw).activate
+  end
+
+  # Initialize a new DrawActivator
+  #
+  # @attr draw [Draw] the draw in question
+  # @attr mailer [ActionMailer::Base] mailer class for sending invitation
+  #   e-mails
+  def initialize(draw:, mailer: StudentMailer)
+    @draw = draw
+    @errors = []
+    @mailer = mailer
+  end
+
+  # Activate a Draw
+  #
+  # @return [Hash{Symbol=>ApplicationRecord,Hash}] A results hash with the
+  #   message to set in the flash and either `nil` or the created object.
+  def activate
+    validate
+    activate_draw if currently_valid?
+    send_emails if currently_valid?
+    currently_valid? ? success : error
+  end
+
+  private
+
+  attr_accessor :draw, :errors, :mailer
+
+  def validate
+    errors << 'Draw must be a draft.' unless draw.draft?
+    errors << 'Draw must have at least one student.' unless draw.students?
+    return if draw.enough_beds?
+    errors << 'Draw must have at least one bed per student; you should add '\
+        'more suites or remove some students.'
+  end
+
+  def currently_valid?
+    errors.empty?
+  end
+
+  def activate_draw
+    return if draw.update(status: 'pre_lottery')
+    errors << 'Draw update failed.'
+  end
+
+  def send_emails
+    draw.students.each do |student|
+      mailer.draw_invitation(student).deliver_later
+    end
+  end
+
+  def success
+    { object: draw, msg: { notice: 'Draw successfully initiated.' } }
+  end
+
+  def error
+    {
+      object: nil,
+      msg: { error: "There was a problem initiating the draft:\n#{error_msgs}" }
+    }
+  end
+
+  def error_msgs
+    errors.join("\n")
+  end
+end
--- a/app/views/draws/show.html.erb
+++ b/app/views/draws/show.html.erb
@@ -11,6 +11,12 @@
       <% end %>
     </tbody>
   </table>
+  <% if @draw.pre_lottery? && @draw.intent_deadline %>
+    <p><%= intent_deadline_str(@draw) %></p>
+  <% end %>
 </div>
+<% if policy(@draw).activate? %>
+  <%= link_to 'Begin draw process', activate_draw_path(@draw), method: :patch %>
+<% end %>
 <%= link_to 'Delete', draw_path(@draw), method: :delete %>
 
--- a/app/views/student_mailer/draw_invitation.html.erb
+++ b/app/views/student_mailer/draw_invitation.html.erb
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta content='text/html; charset=UTF-8' http-equiv='Content-Type' />
+  </head>
+  <body>
+    <p>Hi <%= @user.name %>,</p>
+    <p>The housing process for <%= @res_college.name %> is now open. Please log in to <a href="<%= @res_college.vesta_url %>">Vesta</a> to indicate whether you'll be living on- or off-campus by <%= @intent_deadline %>. If you'll be living on-campus you'll be able to view available housing and form or join a housing group.</p>
+    <p>Please e-mail <%= @res_college.admin_email %> if you have any questions.</p>
+    <p>Sincerely,<br /><%= @res_college.dean %></p>
+  </body>
+</html>
--- a/spec/features/draws/draw_start_spec.rb
+++ b/spec/features/draws/draw_start_spec.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.feature 'Draw activation' do
+  let(:draw) { FactoryGirl.create(:draw_with_members, status: 'draft') }
+  context 'as admin' do
+    before { log_in(FactoryGirl.create(:admin)) }
+
+    it 'can be initiated' do
+      visit draw_path(draw)
+      click_link('Begin draw process')
+      expect(page).to have_css('.flash-notice',
+                               text: 'Draw successfully initiated.')
+    end
+
+    it 'redirects on failure' do
+      draw = FactoryGirl.create(:draw, status: 'draft')
+      visit draw_path(draw)
+      click_link('Begin draw process')
+      expect(page).to have_css('.flash-error')
+    end
+  end
+
+  context 'as student' do
+    before { log_in(FactoryGirl.create(:student)) }
+
+    it 'cannot see the start button' do
+      visit draw_path(draw)
+      expect(page).not_to have_link('Begin draw process')
+    end
+  end
+end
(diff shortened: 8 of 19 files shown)


app/services/draw_starter.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+#
+# Service object to handle the starting of a draw. Checks to make sure the draw
+# is in the correct status, updates the status, and sends e-mail invitations to
+# students in the draw.
+class DrawStarter
+  # Class method to permit calling :start on the class without instantiating the
+  # service object directly
+  #
+  # @attr draw [Draw] the draw in question
```
esoterik: I think we should just enforce instantiation or make this a module instead of a class

app/views/draws/show.html.erb
```diff
@@ -12,5 +12,8 @@
     </tbody>
   </table>
 </div>
+<% if can_start_draw?(draw: @draw, policy: policy(@draw)) %>
```
esoterik: `policy(@draw).start?` should work given my suggested change to the policy above

spec/features/draws/draw_start_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  context 'as student' do
+    before { log_in(FactoryGirl.create(:student)) }
+
+    it 'cannot see the start button' do
+      visit draw_path(draw)
+
+      expect(page).not_to have_link('Begin draw process')
+    end
+  end
```
esoterik: I'm hesitant to leave these permission-based feature specs long-term. I think they're good when behavior differs greatly between roles, but might be overkill for things like seeing a button. 

Perhaps we want to have some view specs instead?

app/policies/draw_policy.rb
```diff
@@ -9,6 +9,10 @@ def update?
     user.rep? || user.admin?
   end
 
+  def start?
+    user.admin?
```
esoterik: changing this to `user.admin? && record.draft?` eliminates the need for `DrawHelper`

spec/services/draw_starter_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+      draw = double('draw')
+      draw_starter = create_mock_draw_starter(draw: draw)
+      DrawStarter.start(draw: draw)
+
+      expect(draw_starter).to have_received(:start)
+    end
+  end
+
+  def create_mock_draw_starter(param_hash)
+    draw_starter = double('draw_starter')
```
esoterik: I *think* that `instance_spy` is generally the preferred mock method ([thoughtbot post](https://robots.thoughtbot.com/a-closer-look-at-test-spies)), though there are trade-offs. 

If this is kept as a standard mock, though, it should be a [verifying double](https://relishapp.com/rspec/rspec-mocks/v/3-5/docs/verifying-doubles) so it checks that the stubbed methods are present on the actual class/instance of said class

app/controllers/draws_controller.rb
```diff
@@ -31,6 +33,11 @@ def destroy
     handle_action(**result)
   end
 
+  def start
```
esoterik: this is super nit-picky but I feel like `start` is too ambiguous between opening up the draw to students and running the lottery -- `open` might be better for this transition?

app/helpers/draw_helper.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+#
+# Helper module for Draws
+module DrawHelper
+  # Check to see if a draw is startable by the current user. Returns true if the
+  # draw is in the 'draft' status and the user satisfies the Pundit policy,
+  # false otherwise.
```
orenyk: Ha, thanks for catching that. I forgot we could do more than assert on `user` in policies 😛. I'll update.

app/services/draw_starter.rb
orenyk: I actually prefer this paradigm for service objects and I'm surprised that I forgot to mention it when we started implementing them. I picked it up from the Upcase forums when I was doing some of the Ruby challenges: [this]() is the thread and [here](http://blog.codeclimate.com/blog/2012/11/14/why-ruby-class-methods-resist-refactoring/)'s a great article on it. Take a look and see if it changes your mind; the reason why we wouldn't want a module is because modules don't maintain state, making refactoring and method extraction difficult, and instantiation just feels unnecessary for single-use objects like this. I really think we should stick with it moving forward and potentially even refactor our controllers and existing service objects to follow that model.

app/views/draws/show.html.erb
orenyk: yea it's what I originally had and then I starting thinking about `@draw.status` and for some reason forgot that the policy could include that as well 😕.

app/policies/draw_policy.rb
orenyk: Way to keep hammering the point home :-).

spec/services/draw_starter_spec.rb
orenyk: Good catch, I wasn't 100% sure why we kept using instance methods so figured I'd keep it simple for now, thanks for the reference.

orenyk: Yea after reading the post I'm convinced. I'll switch them everywhere in the tests.

app/controllers/draws_controller.rb
orenyk: Agreed; especially since starting the draw could mean starting the actual draw / lottery phase (this was already bothering me). How about `initiate`? `open` doesn't really work for me - I get that the implication is "open for business," so to speak, but it feels like weird verb to use `DrawOpener` vs `DrawInitiator`. Thoughts?

spec/features/draws/draw_start_spec.rb
orenyk: Hm, maybe? I don't have a lot of experience with View specs in Rails; I've done a bit of that in Phoenix but there everything is basically unit testable so it's really easy. I'll take a look and see about moving everything around.

spec/services/draw_activator_spec.rb
```diff
@@ -0,0 +1,91 @@
+# frozen_string_literal: true
```
esoterik: this is missing `require rails_helper`

spec/services/draw_activator_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+  # 'pre-lottery' (dashes are problematic in Ruby symbols) - but there is no
+  # verification of the fact that the `:pre_lottery?` method is a valid method
+  # on the Draw class, e.g. that the 'pre_lottery' value is a valid enum value.
+  # This kinda makes sense since a lot of this is accomplished through
+  # `method_missing`, but it's def something to keep in mind. The problem is
+  # that we can write a test that would pass against broken code since it
+  # doesn't check for enum validity. I tried switching to FactoryGirl's
+  # `build_stubbed` but I ran into major issues with data being persisted or
+  # being unable to generate a valid draw for the service object without
+  # persisting all the things. Thoughts?
```
esoterik: re: verifying doubles not verifying:
I tested this a bit locally and I'm surprised because I didn't run into any of this when refactoring specs for reservations. I did a small amount of research and you're right, it has to do with the enum methods being [dynamically defined](https://relishapp.com/rspec/rspec-mocks/v/3-5/docs/verifying-doubles/dynamic-classes).

It does mean that we're opening ourselves up to some risk when mocking, but I think it's worth it overall to not persist data for testing method details. We'll still have integration tests to cover us on major behaviors, and I don't think there's a better solution for verifying the mocks.

re: issues with `build_stubbed`:
ActiveRecord associations do not play well when some of your data is mocked and some of it is real, I know that pain. Here I don't think we'll have to deal with it (I think I fixed the persisted spec above, though I could be wrong), but long-term we might want to pull over my `Mocker` class from reservations to make dealing with associations easier when we don't want to persist data.

spec/services/draw_activator_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+
+    it 'updates the status of the draw to pre_lottery' do
+      # draw = create_mock_draw
+      draw = create_draw_data
+
+      result = described_class.activate(draw: draw)
+
+      expect(result[:obj]).to be_pre_lottery
+    end
```
esoterik: I was able to get the following to pass without persistence:

      draw = instance_spy('Draw', draft?: true, has_students?: true,
                                  has_enough_beds?: true)

      described_class.activate(draw: draw)

      expect(draw).to have_received(:update_attributes)
        .with(status: 'pre_lottery')

Which I think tests what we want it to?

spec/services/draw_activator_spec.rb
orenyk: Ha good catch. That's the second time I did that 😕

spec/services/draw_activator_spec.rb
orenyk: Yea..... I don't think it solves my problem (the enum could change and this spec would still pass), but that may be fine as long as we have at least one integration test that _would_ fail if the enum changed, which is what we want. I was implicitly testing this by stubbing `update_attributes(status: 'pre_lottert')` to return the updated spy with `pre_lottery?: true` stubbed, but your assertion is better as we're really testing the message sent out of the service object, not the outcome.

That said, should we? I feel like we should only send the invitation e-mails if that update is successful, so as a sanity check should we make that action dependent on the result of `update_attributes`?

spec/services/draw_activator_spec.rb
orenyk: Thanks for confirming my suspicions and validating that I wasn't being entirely stupid. I'm open to implementing the `Mocker` again if we feel we need it - I think we've been ok without the additional complexity and just manually dealing with `instance_spy`s, but we can come back to this if it comes up.

spec/services/draw_activator_spec.rb
orenyk: Wait, if that's the case, why is this test even running? Do we even need it?

orenyk:
@esoterik ok I think this is ready!

orenyk:
This is going to fail rubocop, btw, I'll fix those after the build.

orenyk:
I'll also check out codeclimate in a bit.

orenyk:
The only CodeClimate failures at this point should be some `TODO` comments in the code - can we disable that check? I'm kind of a fan of leaving notes to our future dev selves in the code where relevant. Let me know what you think!

spec/services/draw_activator_spec.rb
esoterik: it configures DatabaseCleaner and other things, so we need it

app/controllers/draws_controller.rb
```diff
@@ -42,10 +48,18 @@ def authorize!
...
   def set_draw
     @draw = Draw.find(params[:id])
   end
+
+  # TODO: I don't love that this is necessary for #activate but it feels like
+  # the 'show' template is the right one to render; however that view requires
+  # the intent metrics to render properly.
+  def calculate_metrics
+    @intent_metrics = IntentMetricsQuery.call(@draw)
+  end
```
esoterik: I think that this is fine

app/mailers/student_mailer.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+#
+# Mailer class for user e-mails
+class StudentMailer < ApplicationMailer
+  default from: 'no-reply@vesta.app' # SHOULD WE HAVE AN ADMIN ADDRESS?
```
esoterik: we should probably have an admin address; it will probably make sense for students to want to respond to emails individually in some circumstances.

esoterik: this can be a separate issue at this point I think

app/mailers/student_mailer.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  def draw_invitation(user)
+    @user = user
+    @intent_deadline = user.draw.intent_deadline
+    # TODO: figure this out - singleton class, e.g. app config, or do we add a
+    # class for now that we later use for multitenancy?
```
esoterik: I think it should end up as an app config model; it'll work better when we have multi-tenancy

app/models/draw.rb
```diff
@@ -2,16 +2,56 @@
...
+  #   has students
+  def enough_beds?
+    bed_count >= student_count
+  end
+
+  private
+
+  # TODO: I had originally made this public but now I'm not sure it needs to be.
+  # Let me know if you'd like to keep it private or not and we can remove the
+  # YARD docs and the spec.
```
esoterik: I think its fine for these to be private

app/services/draw_activator.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  # Activate a Draw
+  #
+  # @return [Hash{Symbol=>ApplicationRecord,Hash}] A results hash with the
+  #   message to set in the flash and either `nil` or the created object.
+  def activate
+    validate
+    activate_draw if valid?
+    send_emails if valid?
+    valid? ? success : error
```
esoterik: ~~I think it makes sense to wrap this all in one `if`~~

hmm, this isn't the biggest problem, but it's unclear that validity can change between the steps of this method. Not quite sure what to suggest, and it's also fine as it is -- just something I thought while reading through the class

app/views/student_mailer/draw_invitation.html.erb
```diff
@@ -0,0 +1,13 @@
+<% #TODO: figure out how to do nice e-mail templates %>
```
esoterik: should this be a new issue?

spec/features/draws/draw_start_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      visit draw_path(draw)
+      click_link('Begin draw process')
+      expect(page).to have_css('.flash-notice',
+                               text: 'Draw successfully initiated.')
+    end
+
+    it 'redirects to show on failure' do
+      visit draw_path(draw)
+      click_link('Begin draw process')
+      # TODO: figure out if there's a better assertion?
```
esoterik: this is fine I think

spec/features/draws/draw_start_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      expect(page).to have_css('.flash-error')
+    end
+
+    def create_draw_resources(draw)
+      FactoryGirl.create_list(:student, 2, draw: draw)
+      draw.suites << FactoryGirl.create(:suite_with_rooms, rooms_count: 2)
+    end
```
esoterik: there's a `:draw_with_members` factory that you can use for this!

spec/features/draws/draw_start_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  context 'as student' do
+    before { log_in(FactoryGirl.create(:student)) }
+
+    it 'cannot see the start button' do
+      visit draw_path(draw)
+      expect(page).not_to have_link('Begin draw process')
+    end
+  end
```
esoterik: I still think we're better off using [view specs](https://robots.thoughtbot.com/how-we-test-rails-applications#view-specs) for these kinds of things

spec/models/draw_spec.rb
```diff
@@ -16,4 +17,47 @@
       expect(draw.suite_sizes).to match_array([1, 2, 3])
     end
   end
+
+  # TODO: should we keep this? I originally wrote this before I decided to make
+  # #student_count private, but we can make it public again if we think it's
+  # worth it.
+  #
```
esoterik: might as well keep it, it doesn't persist data so it's fast, and we can always delete it later

spec/services/draw_activator_spec.rb
```diff
@@ -0,0 +1,95 @@
+# frozen_string_literal: true
+require 'spec_helper'
```
esoterik: notably our `spec_helper` and `rails_helper` configurations are different -- I've been using `rails_helper` everywhere because of the DatabaseCleaner config

spec/services/draw_activator_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  def create_mock_draw_activator(param_hash)
+    instance_spy('draw_activator').tap do |draw_activator|
+      allow(described_class).to receive(:new).with(param_hash)
+        .and_return(draw_activator)
+    end
+  end
+
+  def create_valid_mock_draw_with_students(num_students: 2)
+    students = Array.new(num_students) { |_n| instance_spy('user') }
```
esoterik: The block argument isn't necessary here

spec/services/draw_activator_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+    end
+
+    it 'sets the object key to nil in the hash on failure' do
+      draw = instance_spy('draw', validity_stubs(valid: false))
+      result = described_class.activate(draw: draw)
+      expect(result[:object]).to be_nil
+    end
+  end
+
+  def create_mock_draw_activator(param_hash)
```
esoterik: I think that the `create_` prefix can be dropped from these method names

esoterik:
re: codeclimate erroring on the TODOs: 

I think we should keep the checks because otherwise TODOs end up getting lost in the codebase, but we can keep some of them in place.

app/mailers/student_mailer.rb
orenyk: Yea, that's what I was thinking.

app/mailers/student_mailer.rb
orenyk: Yea exactly. I think that some kind of high-level model for the root-level user (similar to `Department` in Shifts) will work fine.

app/models/draw.rb
orenyk: Ok, should I remove the documentation? I'll definitely remove the test.

app/services/draw_activator.rb
orenyk: changing the method name to `currently_valid?`

spec/services/draw_activator_spec.rb
orenyk: Yea good point.

app/controllers/draws_controller.rb
orenyk: 👍

app/mailers/student_mailer.rb
orenyk: I also think it can be replaced with our eventual config model, so I'll leave a comment to that effect.

app/views/student_mailer/draw_invitation.html.erb
orenyk: #100

spec/features/draws/draw_start_spec.rb
orenyk: 👍

spec/features/draws/draw_start_spec.rb
orenyk: awesome! I still wanted to use an invalid draw for one spec, but I think it's ok as is.

spec/features/draws/draw_start_spec.rb
orenyk: I'll open an issue

spec/services/draw_activator_spec.rb
orenyk: Should DatabaseCleaner be in `spec_helper` since it's not Rails specific?

spec/models/draw_spec.rb
orenyk: I had to modify the expectation since RSpec fails if you call a private method directly, so instead I used `send`. I also left a comment that it should be removed if it ever fails.

spec/services/draw_activator_spec.rb
orenyk: 👍

spec/services/draw_activator_spec.rb
orenyk: good catch!

orenyk:
@esoterik ok I removed all but the one TODO we discussed and addressed the other issues. My only last concern is the `rails_helper` vs `spec_helper` question - see my note above about DatabaseCleaner. I'm fine to just use `rails_helper` everywhere, but I've read articles suggesting not to ([here](https://robots.thoughtbot.com/document-explicit-dependencies-through-tests)'s one example). Let me know what you think.

Also squashed 😄. Thanks!

spec/services/draw_activator_spec.rb
esoterik: hmm. I think that any tests that persist data will be Rails-specific, so it's probably fine how it is?

esoterik:
this is good to go pending resolution of the DatabaseCleaner config

PR #105, opened 2017-01-20
[93] Add special (non-draw) group creation

### LINKED ISSUE #93: Admins can manually create groups outside of draws
This is for exceptional case handling outside of draws, we want them to be able to create ad-hoc groups of students and assign said groups to suites. We can use our existing data models as long as we remove the `validate :draw, presence: true` from groups and then everything else should be fine.

orenyk:
Resolves #93
- Add root-level groups resource handled by SpecialGroupsController
- Add SpecialGroup wrapper/delegator class so that we can use different permissions for the SpecialGroupsController
- Add SuiteSizesQuery to get all unique suite sizes in a given relation
- Add SpecialGroupCreator service object to ignore draw logic
- Split up groups _form partial for reuse in special_groups _form


app/models/group.rb
```diff
@@ -10,18 +10,17 @@
 class Group < ApplicationRecord
   belongs_to :leader, class_name: 'User'
   belongs_to :draw
-  has_many :memberships
+  has_many :memberships, dependent: :delete_all
```
orenyk: Should we test this somehow? It's pretty lucky that I caught it by just playing with the app manually... a good reminder that automated tests don't always cover every aspect of the app. Where would this test go? On the model? e.g.

```ruby
it 'destroys dependent memberships` do
  group = FactoryGirl.create(:special_group)
  memberships = group.memberships
  group.destroy
  expect(Membership.find(memberships.map(&:id))).to be_empty
end
```

I guess that could work, let me know what you think and I'll add it (and we can do something similar for users).

orenyk:
@esoterik finally ready! It's a bit of a doozy, though, but hopefully I've covered all the edge cases. ~~See my note about testing `dependent: :delete` for our membership associations, I can add it pretty easily.~~ I'm just going to add it and re-push.

app/models/group.rb
orenyk: Nvm, I added these specs to the model specs for both user and group.

app/models/user.rb
```diff
@@ -7,7 +7,7 @@ class User < ApplicationRecord
          :recoverable, :rememberable, :trackable, :validatable
 
   belongs_to :draw
-  has_one :membership
+  has_one :membership, dependent: :destroy
```
orenyk: Note that we can use `dependent: :destroy` here since the group isn't destroyed when a user is hypothetically destroyed. This will ensure that all the membership / group callbacks are correctly called.

app/models/group.rb
orenyk: Note the use of `delete_all` due to callback loops when `destroy` was used (group is being destroyed, it calls destroy on the membership, which attempts to update the group and it gets messy and broken).

app/controllers/special_groups_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def destroy
+    result = Destroyer.new(object: @group, name_method: :name).destroy
+    handle_action(**result)
+  end
+
+  private
+
+  def authorize!
+    @group ? authorize(SpecialGroup.new(@group)) : authorize(SpecialGroup)
```
esoterik: do you prefer the ternary syntax for these?

app/controllers/special_groups_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+  def special_group_params
+    params.require(:group).permit(:size, :leader_id, member_ids: [])
+  end
+
+  def set_group
+    @group = Group.find(params[:id])
+  end
+
+  def set_form_data
+    @group ||= Group.new
```
esoterik: this is nice!!

app/models/group.rb
```diff
@@ -47,6 +46,16 @@ def add_leader_to_members
...
 
+  def validate_suite_size_inclusion
+    if draw.present?
+      return if draw.suite_sizes.include? size
+      errors.add :size, 'must be a suite size included in the draw'
+    else
+      return if SuiteSizesQuery.call.include? size
+      errors.add :size, 'must be a valid suite size'
+    end
+  end
```
esoterik: if we end up with more draw presence conditionals, we should consider creating a `DrawlessGroup` object to contain all of this behavior, but this is fine for now

esoterik: (this is maybe worth commenting somewhere?)

app/controllers/special_groups_controller.rb
orenyk: It's more concise, so I figured why not. What do you think?

app/controllers/special_groups_controller.rb
orenyk: It was born of necessity 😄. Rubocop / code smells for the win!

app/models/group.rb
orenyk: Yea... excellent catch. I'll add a comment to this validation method (maybe with a TODO header just to keep us on our toes). Let me know what you think!

app/services/creators/special_group_creator.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  def process_params
+    @params = params.to_h.transform_keys(&:to_sym)
+    remove_blank_members if params[:member_ids]
+    remove_members_and_leader_from_draw
+  end
+
+  def remove_blank_members
+    @params[:member_ids] = params[:member_ids].reject(&:empty?)
+  end
```
orenyk: Hmm, the alternative could be just to use GroupCreator and add some `draw.present?` logic? I'll ping you on Slack to discuss.

orenyk:
I'm also going to add some logic to the service object to switch intents to `on_campus`, good catch!

orenyk:
@esoterik how's this? I still haven't updated the `DrawlessGroupsPolicy`, but we can leave it for now and open a new issue if you'd like to adjust those permissions. Let me know!

PR #111, opened 2017-01-22
Seed Script

### LINKED ISSUE #109: Seed Script

esoterik:
Resolves #109 

I ended up removing all of the progress-related things for now until we figure out how to disable them when the test suite is running the generators. 

This is probably good to go.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,5 +1,22 @@
 # frozen_string_literal: true
-# This file should contain all the record creation needed to seed the database
-# with its default values.
-# The data can then be loaded with the rails db:seed command
-# (or created alongside the database with db:setup).
+# rubocop:disable Rails/Output
+
+puts 'Generating seed data....'
+
+Generator.generate_admin(email: 'email@email.com', password: 'passw0rd')
+Generator.generate(model: 'building', count: 2)
+Generator.generate(model: 'suite', count: 15)
+
+# generate bedrooms
+Generator.generate(model: 'room', count: 30)
+# generate common rooms
+Generator.generate(model: 'room', count: 10, beds: 0)
+
+Generator.generate(model: 'user', count: 15)
+
+# fix this eventually so that we never generate empty suites
+Suite.where(size: 0).destroy_all
+
+Generator.generate(model: 'draw', count: 3)
+
+puts 'Finished!'
--- a/lib/seed/draw_generator.rb
+++ b/lib/seed/draw_generator.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+# Seed script generator for Draws
+class DrawGenerator
+  def self.generate(**overrides)
+    new(overrides: overrides).generate
+  end
+
+  def initialize(overrides: {})
+    gen_params(overrides: overrides)
+  end
+
+  def generate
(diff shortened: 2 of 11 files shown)


esoterik:
Code climate is failing but I think it's okay

lib/seed/draw_generator.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  private
+
+  attr_reader :params
+
+  def gen_params(overrides: {})
+    check_member_existence
+    @params ||= { suites: Suite.all.sample(3),
+                  students: User.where(role: 'student').sample(5),
```
esoterik: this needs to also only get students not in a draw

lib/seed/suite_generator.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+# Seed script generator for Suites
+class SuiteGenerator
+  def self.generate(**overrides)
+    new(overrides: overrides).generate
+  end
+
+  def self.generate_admin(**overrides)
+    new(overrides: overrides.merge(role: 'admin')).generate
+  end
```
esoterik: forgot to get rid of these

db/seeds.rb
```diff
@@ -1,5 +1,22 @@
...
+
+# generate bedrooms
+Generator.generate(model: 'room', count: 30)
+# generate common rooms
+Generator.generate(model: 'room', count: 10, beds: 0)
+
+Generator.generate(model: 'user', count: 15)
+
+# fix this eventually so that we never generate empty suites
+Suite.where(size: 0).destroy_all
```
orenyk: I feel like the only way to avoid this is by somehow generating suites with rooms in the `SuiteGenerator` itself. I'm kind of ok with that, tbh.

lib/seed/draw_generator.rb
```diff
@@ -0,0 +1,31 @@
...
+    check_member_existence
+    @params ||= { suites: Suite.all.sample(3),
+                  students: User.where(role: 'student', draw_id: nil).sample(5),
+                  name: "#{FFaker::Music.artist} Draw" }.merge(overrides)
+  end
+
+  def check_member_existence
+    5.times { SuiteGenerator.generate } unless Suite.all.count >= 5
+    8.times { UserGenerator.generate } unless User.all.count >= 8
+  end
```
orenyk: Were these numbers chosen arbitrarily?

spec/lib/generator_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    it klass.to_s do
+      expect { described_class.generate(model: klass.to_s) }.to \
+        change { klass.count }.by(1)
+    end
+  end
+
+  OBJECTS.each { |o| it_behaves_like 'generates a valid', o }
+
+  it 'can generate an admin' do
+    expect { described_class.generate_admin }.to change { User.count }.by(1)
```
orenyk: Should we check that the user created is an admin? That would just change the assertion to `change { User.where(role: 'admin').count }`.

lib/seed/draw_generator.rb
esoterik: yeah pretty much

esoterik:
merging!

Issue #107, opened 2017-01-20
Students can Accept Group Invitations


esoterik:
blocked on #90

PR #121, opened 2017-01-27
[56] Request to join groups

### LINKED ISSUE #56: Students Can Request to Join Groups

esoterik:
Resolves #56

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -1,7 +1,8 @@
 # frozen_string_literal: true
 # Controller for Groups
 class GroupsController < ApplicationController
-  prepend_before_action :set_group, only: %i(show edit update destroy)
+  prepend_before_action :set_group, only: %i(show edit update destroy
+                                             request_to_join accept_request)
   before_action :set_draw
   before_action :set_form_data, only: %i(new edit)
 
@@ -31,6 +32,19 @@ class GroupsController < ApplicationController
   def destroy
     result = Destroyer.new(object: @group, name_method: :name).destroy
     handle_action(**result)
+  end
+
+  def request_to_join
+    result = MembershipCreator.create!(group: @group, user: current_user,
+                                       status: 'requested')
+    handle_action(path: draw_group_path(@draw, @group), **result)
+  end
+
+  def accept_request
+    user = User.includes(:membership).find(params['user_id'])
+    result = MembershipUpdater.update(membership: user.membership,
+                                      params: { status: 'accepted' })
+    handle_action(path: draw_group_path(@draw, @group), **result)
   end
 
   private
--- a/app/models/membership.rb
+++ b/app/models/membership.rb
@@ -9,12 +9,17 @@
 #
 # @attr group [Group] The group of the membership.
 # @attr user [User] The user of the membership.
+# @attr status [Integer] Enum for membership status.
+#   (requested, invited, accepted)
 class Membership < ApplicationRecord
-  belongs_to :group, counter_cache: true
+  belongs_to :group
   belongs_to :user
+
+  enum status: %w(accepted invited requested)
 
   validates :group, presence: true
   validates :user, presence: true
+  validates :status, presence: true
   validate :matching_draw, if: ->(m) { m.user.present? && m.group.present? }
   validate :user_on_campus, if: ->(m) { m.user.present? }
   validate :group_is_open, if: ->(m) { m.group.present? }, on: :create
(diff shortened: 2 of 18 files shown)


app/models/membership.rb
```diff
@@ -53,4 +72,17 @@ def user_on_campus
     return if user.on_campus?
     errors.add :user, 'must be living on campus to join a group'
   end
+
+  def update_counter_cache
+    return unless status_changed?
+    increment_counter_cache
```
orenyk: Hmm shouldn't this actually be `if accepted?`? I'm just thinking about cases where it goes from `requested` to `invited` hypothetically - that shouldn't trigger an increment of the counter cache. Thoughts?

app/services/creators/membership_creator.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+#
+# Service object to create memberships.
+class MembershipCreator < Creator
+  # Class method to permit calling :create! on the class without instantiating
+  # the service object directly
+  #
+  # @param [#to_h] params The params object from the MembershipsController.
```
orenyk: Let's fix this comment since there is no `MembershipsController`.

app/services/creators/membership_creator.rb
```diff
@@ -0,0 +1,27 @@
...
+  def self.create!(params)
+    new(params).create!
+  end
+
+  # Initialize a MembershipCreator
+  #
+  # @param [#to_h] params The params object from the MembershipsController.
```
orenyk: Same here re:comment

app/services/updaters/membership_updater.rb
```diff
@@ -0,0 +1,27 @@
...
+  # @param params [#to_h] The new attributes
+  def initialize(membership:, params:)
+    super(object: membership, name_method: nil, params: params)
+  end
+
+  private
+
+  def success
+    { object: [object.group.draw, object.group],
+      msg: { success: "#{object.user.full_name} accepted." } }
```
orenyk: Just  a note, but we may want to make this message configurable for this object since it will likely be different for accepting invitations vs accepting requests. That will likely come into play in #107.

app/views/groups/show.html.erb
```diff
@@ -1,2 +1,11 @@
...
+<% if policy(@group).accept_request? %>
+  <h2>Membership Requests</h2>
+  <% @group.requests.each do |u| %>
+    <%= link_to "Accept #{u.full_name}", draw_group_accept_request_path(@draw, @group, params: { 'user_id' => u.id }), method: :put %>
+  <% end %>
+<% end %>
+<% if policy(@group).request_to_join? %>
+  <%= link_to 'Request To Join', draw_group_request_path(@draw, @group), method: :post %>
+<% end %>
+<%= link_to('Delete', draw_group_path(@draw, @group), method: :delete) if policy(@group).destroy? %>
```
orenyk: Another note, but this feels like it will be worth extracting into a partial for the footer where we can pass in the object in question as a local variable, since `destroy?` is always the relevant action across all the policies. I'll open an issue for this.

spec/features/groups/request_to_join_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      click_on 'Request To Join'
+      expect(page).to have_content('Please review the errors below:')
+    end
+  end
+
+  context 'approving requests' do
+    it 'succeeds' do # rubocop:disable RSpec/ExampleLength
+      group = FactoryGirl.create(:open_group)
+      user = FactoryGirl.create(:student, intent: 'on_campus', draw: group.draw)
+      MembershipCreator.create!(group: group, user: user, status: 'requested')
```
orenyk: Should we just use `Membership.create` here? I'm not sure what the service object adds since we're passing all the same parameters in.

spec/models/group_spec.rb
```diff
@@ -82,4 +83,13 @@
       expect(group.name).to include(leader.name)
     end
   end
+
+  describe '#requests' do
+    it 'returns an array of users who have requested to join' do
+      group = FactoryGirl.create(:open_group)
+      user = FactoryGirl.create(:student, intent: 'on_campus', draw: group.draw)
+      Membership.create(group: group, user: user, status: 'requested')
```
orenyk: Lol, see my [previous comment](https://github.com/YaleSTC/vesta/pull/121/files#r98221781) - we weren't even consistent last night 😛.

spec/features/groups/request_to_join_spec.rb
orenyk: Yea that's what we did in all the other specs, so I'd recommend changing it here.

app/models/membership.rb
esoterik: the increment only works if the status is accepted

app/services/updaters/membership_updater.rb
esoterik: We could just change it to "joined group" instead of "accepted"?

app/controllers/groups_controller.rb
```diff
@@ -33,6 +34,19 @@ def destroy
     handle_action(**result)
   end
 
+  def request_to_join
+    result = MembershipCreator.new(group: @group, user: current_user,
+                                   status: 'requested').create!
```
esoterik: should be changed to just call `create!` directly

PR #125, opened 2017-01-28
115 group new page

spaceraccoon:
Styled it to look like the whiteboard drawing. Simpleform doesn't have a built in inline checkbox for some reason (i.e. checkbox and labels on the same row vs. each option on a new line), so had to use a css override. Essentially, simpleform places each new option in a new "checkbox" div, which occupies a whole line. So I overrode this by making the checkbox css inline and 20% width.

### DIFF
--- a/_inline_checkboxes_override.scss
+++ b/_inline_checkboxes_override.scss
@@ -1,0 +1,9 @@
+.inline-checkbox {
+  display: inline-block;
+  width: 20%;
+  word-wrap: break-word;
+
+  @include breakpoint(medium) {
+    width: 100%;
+  }
+}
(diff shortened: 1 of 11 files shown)


spaceraccoon:
Passes all checks and can be pulled!

esoterik:
@spaceraccoon please add "needs review" after you update a PR

app/assets/stylesheets/inline_checkboxes_override.scss
```diff
@@ -0,0 +1,5 @@
+.inline-checkbox {
+	display: inline-block;
+	width: 20%;
```
orenyk: This totally breaks on mobile, we should have some kind of breakpoint here and probably just increase them to full width.

orenyk:
We'll also have to rebase.

spaceraccoon:
Tried rebasing but there are some weird conflicts? @esoterik @orenyk

spaceraccoon:
@orenyk @esoterik rebased and ready for review!

PR #132, opened 2017-01-30
[127] Add admin creation of users

### LINKED ISSUE #127: Admin can create other admins
Step 1: enter username
Step 2 (future): check for IDR
Step 2 (now): enter profile data
Step 3: profit

## Still to do:

* [x] Write tests for User.cas_auth?
* [ ] Finish implementing UserCreator (with specs) - needs to handle CAS auth and generate random password for non-CAS cases (maybe display in flash for now and in the future branch out after build and send an invite e-mail using Devise's built-in invitation functionality)
* [x] Finish writing commit message
* [x] Add hidden input field on new form for e-mail to be sent through params if disabled
* [x] Send an e-mail to newly created user at end of process via UserCreator (more specs, woo)
* [x] Finish documenting all public methods in all service objects, etc

orenyk:
Resolves #127
- Add two-step user creation process, first a build step where either a username or email (CAS or non-CAS) is passed in and checked for uniqueness, after which the rest of the profile data is created and passed to a more standard Creator-type service object
- Add UserMailer with account confirmation e-mail that optionally includes an auto-generated password for non-CAS auth

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,13 +1,22 @@
 # frozen_string_literal: true
 # User model class for all user types.
 class User < ApplicationRecord
+  # Determine whether or not CAS authentication is being used, must be at the
+  # top of the class to be used in the Devise loading conditional below.
+  #
+  # @return [Boolean] true if the CAS_BASE_URL environment variable is set,
+  #   false otherwise
+  def self.cas_auth?
+    env? 'CAS_BASE_URL'
+  end
+
   # Include default devise modules. Others available are:
   # :confirmable, :lockable, :timeoutable and :omniauthable
-  if env? 'CAS_BASE_URL'
+  if cas_auth?
     devise :cas_authenticatable, :trackable
   else
-    devise :database_authenticatable, :registerable,
-           :recoverable, :rememberable, :trackable, :validatable
+    devise :database_authenticatable, :recoverable, :rememberable, :trackable,
+           :validatable
   end
 
   belongs_to :draw
@@ -50,6 +59,6 @@ class User < ApplicationRecord
   end
 
   def cas_auth?
-    env? 'CAS_BASE_URL'
+    User.cas_auth?
   end
 end
--- a/app/services/user_builder.rb
+++ b/app/services/user_builder.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+#
+# Service object for building a user record for creation. Currently takes in a
+# username, ultimately will request data from IDR.
+class UserBuilder
+  # Allow for the calling of :build on the parent class
+  def self.build(id_attr:)
+    new(id_attr: id_attr).build
+  end
+
+  # Initialize a UserBuilder
+  #
+  # @param id_attr [String] the value for the ID attribute, either username
+  # (CAS) or e-mail (non-CAS)
+  def initialize(id_attr:)
+    @id_attr = id_attr
+    @user = User.new
+    @id_symbol = User.cas_auth? ? :username : :email
+  end
+
+  # Build a user record based on the given input, ensuring that it is unique
+  #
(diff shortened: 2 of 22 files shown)


spec/services/creators/user_creator_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+          result = described_class.create!(params)
+          expect(result[:object].password).not_to be_empty
+        end
+      end
+      # TODO: this test currently fails because stubbing out User.cas_auth?
+      # doesn't change how the User class was loaded (without CAS), and so it's
+      # still running password validations. Even switching to any_instance_of
+      # doesn't help because it has to do with when Rails loads the User
+      # class... not sure what to do.
+      context 'with CAS' do # rubocop:disable RSpec/NestedGroups
```
orenyk: I'm not sure how to make this test pass, it may not be possible since we'd have to reload the User class. Thoughts?

app/models/user.rb
```diff
@@ -1,13 +1,21 @@
 # frozen_string_literal: true
 # User model class for all user types.
 class User < ApplicationRecord
+  # Determine whether or not CAS authentication is being used
+  #
```
esoterik: we should probably note that this must be at the top of the class in order for the devise conditional to work

app/services/user_builder.rb
```diff
@@ -0,0 +1,60 @@
...
+  # (CAS) or e-mail (non-CAS)
+  def initialize(id_attr:)
+    @id_attr = id_attr
+    @user = User.new
+    @id_symbol = User.cas_auth? ? :username : :email
+  end
+
+  # Build a user record based on the given input, ensuring that it is unique
+  #
+  # @return [Hash{symbol=>User,HAsh}] a results hash with a message to set in
```
esoterik: typo ("HAsh")

app/services/user_builder.rb
```diff
@@ -0,0 +1,60 @@
...
+    success
+  end
+
+  private
+
+  attr_accessor :user
+  attr_reader :id_attr, :id_symbol
+
+  def result_hash
+    { object: nil, user: user }
```
esoterik: we might want to document why we set `object: nil`

app/services/user_builder.rb
```diff
@@ -0,0 +1,60 @@
...
+  def initialize(id_attr:)
+    @id_attr = id_attr
+    @user = User.new
+    @id_symbol = User.cas_auth? ? :username : :email
+  end
+
+  # Build a user record based on the given input, ensuring that it is unique
+  #
+  # @return [Hash{symbol=>User,HAsh}] a results hash with a message to set in
+  #   the flash, either nil or the record as the :object value, the user record
```
esoterik: here the object is always nil

app/views/users/new.html.erb
```diff
@@ -0,0 +1,14 @@
+<h1>Add new user</h1>
+<%= simple_form_for @user do |f| %>
+  <%= f.input :email, disabled: !@user.email.empty? %>
+  <%= f.input(:email, as: :hidden) unless @user.email.empty? %>
+  <% unless @user.try(:username).try(:empty?) %>
```
esoterik: ooo, `try` is nice

spec/services/creators/user_creator_spec.rb
esoterik: I think we're going to have to actually use an ENV wrapper here, which is ugly but will work

orenyk:
Approved by @esoterik following Slack chat, we concluded that the failing spec shouldn't block us (an ENV wrapper didn't help) so we're leaving it pending and opening an issue. Merging following green Travis builds.

app/services/user_builder.rb
orenyk: Done in the comments for `build`.

Issue #57, opened 2016-12-02
Groups can be Frozen


esoterik:
Should be able to be performed by the group leader OR any admin.

Blocked on #49

esoterik:
blocked on #81

orenyk:
@esoterik who/what was this blocked on? Was it #90?

esoterik:
If a leader decides to freeze/lock a group, it must be confirmed by all members. This can be accomplished by adding a `locked` boolean to `Membership` which is set when a single member agrees to freeze a group.

Admins should be able to skip the approve process and lock groups whenever they want as long as they are full.

PR #136, opened 2017-02-02
[117] Add suite selection for drawless groups

### LINKED ISSUE #117: Admins can assign special groups to suites
Extension of #93, we'll do just special groups for now since I'm not 100% sure how we should handle groups in draws (should they be removed when assigned to a suite by an admin outside of the draw process, etc).

orenyk:
Resolves #117
- Add SuiteSelector service object to handle suite assignment to groups
- Add select_suite action to DrawlessGroupsController
- Add Suite.available scope

### DIFF
--- a/app/policies/drawless_group_policy.rb
+++ b/app/policies/drawless_group_policy.rb
@@ -2,4 +2,7 @@
 #
 # Policy for permissions on special (non-draw) housing groups
 class DrawlessGroupPolicy < ApplicationPolicy
+  def select_suite?
+    user.admin? && record.locked?
+  end
 end
--- a/app/services/suite_selector.rb
+++ b/app/services/suite_selector.rb
@@ -1,0 +1,84 @@
+# frozen_string_literal: true
+#
+# Service object to select suites for a given group.
+class SuiteSelector
+  # Allow for :select to be called on the parent class
+  def self.select(**params)
+    new(**params).select
+  end
+
+  # Initialize a new SuiteSelector
+  #
+  # @param group [Group] the group to assign the suite to
+  # @param suite_id [Integer] the suite id to assign to the group
+  def initialize(group:, suite_id:)
+    @group = group
+    @suite_id = suite_id
+    @errors = []
+  end
+
+  # Select / assign a suite to a group. Checks to make sure that the grou does
+  # not curretly have a suite assigned, that the suite_id corresponds to an
+  # existing suite, and that the suite is not currently assigned to a different
+  # group.
+  #
+  # @return [Hash{symbol=>Group,Hash}] a results hash with a message to set in
+  #   the flash, nil or the group as the :object, and an action to render.
+  def select
+    if assign_suite_to_group
+      success
+    else
+      error
+    end
+  end
+
+  private
+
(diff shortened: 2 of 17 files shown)


orenyk:
Still waiting on specs for the SuiteSelector. I'm also considering eliminating the validation on the Group not already having a suite assigned, unless we add a separate button to unassign the suite? Let's chat tomorrow.

orenyk:
Ok, this is ready for review but we need to discuss whether or not we want to permit "swapping" suites between groups. Let me know what you think!

app/services/suite_selector.rb
```diff
@@ -0,0 +1,83 @@
...
+  def success
+    {
+      object: group,
+      msg: { success: "Suite #{suite.number} assigned to #{group.name}" }
+    }
+  end
+
+  def error
+    {
+      object: nil, action: 'show',
```
esoterik: I think that the action should explicitly be specified in the controller

app/policies/drawless_group_policy.rb
```diff
@@ -2,4 +2,7 @@
 #
 # Policy for permissions on special (non-draw) housing groups
 class DrawlessGroupPolicy < ApplicationPolicy
+  def select_suite?
+    user.admin? && record.full?
```
esoterik: should be `record.locked?` I think

esoterik: possibly this also means that we should default admin created groups to be always locked

app/services/suite_selector.rb
```diff
@@ -0,0 +1,83 @@
...
+
+  # Select / assign a suite to a group. Checks to make sure that the grou does
+  # not curretly have a suite assigned, that the suite_id corresponds to an
+  # existing suite, and that the suite is not currently assigned to a different
+  # group.
+  #
+  # @return [Hash{symbol=>Group,Hash}] a results hash with a message to set in
+  #   the flash, nil or the group as the :object, and an action to render.
+  def select
+    if valid? && assign_suite_to_group
```
esoterik: I'm not sure that I like having conditionals that also assign attributes. I think I would prefer
```ruby
assign_suite_...
if valid?
...
```
with `valid?` checking if suite assignment was successful

orenyk: We use `if user.save` a lot... is there a reason why this would be different? We can't put the `assign_suite_...` method first since that actually calls `group.update` and the validations aren't database- or model-backed.

app/services/suite_selector.rb
orenyk: Sure

app/policies/drawless_group_policy.rb
orenyk: Do we have `locked` yet? I thought that was part of #57.

orenyk:
Wow this turned into a nightmare when I tried to go through this manually. I opened up some secondary issues and addressed one of your comments with responses to the other two. Let me know what you think @esoterik!

app/services/suite_selector.rb
orenyk: resolved during phone call.

PR #150, opened 2017-02-04
[149] Updated show views for inventory and draws. Added buttons to navbar.

### LINKED ISSUE #149: Add Navbar Buttons
Different buttons for admin/user

spaceraccoon:
Resolves #149. Rough PR. Waiting to clear checks.

### DIFF
--- a/app/views/application/_nav.html.erb
+++ b/app/views/application/_nav.html.erb
@@ -7,6 +7,15 @@
       <strong><%= link_to 'Vesta', root_path %></strong>
     </div>
     <div id="responsive-menu">
+      <div class="top-bar-left">
+        <% if current_user %>
+          <% if current_user.admin? %>
+            <%= render 'nav_admin' %>
+          <% else %>
+             <%= render 'nav_student' %>
+           <% end %>
+        <% end %>
+      </div>
       <div class="top-bar-right">
         <ul class="menu">
           <% if current_user %>
--- a/app/views/buildings/show.html.erb
+++ b/app/views/buildings/show.html.erb
@@ -1,3 +1,9 @@
 <h1 class="building-name"><%= @building.name %></h1>
-<%= link_to 'Delete', building_path(@building), method: :delete %>
-
+<h2>Suites</h2>
+<ul>
+  <% @suites.each do |suite| %>
+    <li><%= link_to suite.number, suite_path(suite) %> - <%= pluralize(suite.size,'bed') %></li>
+  <% end %>
+</ul>
+<p><%= link_to 'Add suite', new_suite_path %></p>
+<p><%= link_to 'Delete', building_path(@building), method: :delete %></p>
--- a/app/views/draws/show.html.erb
+++ b/app/views/draws/show.html.erb
@@ -15,11 +15,14 @@
     <p><%= intent_deadline_str(@draw) %></p>
   <% end %>
   <% if policy(@draw).intent_report? %>
-    <%= link_to 'View intent report', draw_intent_report_path(@draw) %>
+    <p><%= link_to 'View intent report', draw_intent_report_path(@draw) %></p>
   <% end %>
+  <% if policy(@draw).activate? %>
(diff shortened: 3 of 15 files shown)


esoterik:
@spaceraccoon I think you forgot to add untracked files to the diff -- can you update?

orenyk:
@spaceraccoon please update the branch so we can review, thanks!

spaceraccoon:
Eek sorry! Ready @orenyk @esoterik

app/views/application/_nav.html.erb
```diff
@@ -7,6 +7,25 @@
       <strong><%= link_to 'Vesta', root_path %></strong>
     </div>
     <div id="responsive-menu">
+      <div class="top-bar-left">
```
orenyk: There needs to be logic here around what is available to different users / in different states. For example, we should probably only show the `Log In` link when a user is not logged in. Additionally, all of these links feel like admin-only links so they should likewise be restricted to admin users. Does that make sense?

app/views/buildings/show.html.erb
```diff
@@ -1,3 +1,9 @@
 <h1 class="building-name"><%= @building.name %></h1>
+<h2>Suites</h2>
+<ul>
+  <% @suites.each do |suite| %>
+    <li><%= link_to suite.number, suite_path(suite) %> - <%= suite.size %> beds</li>
```
orenyk: We should pluralize `beds` based on the suite size.

app/views/draws/show.html.erb
```diff
@@ -19,7 +19,10 @@
   <% end %>
 </div>
 <% if policy(@draw).activate? %>
-  <%= link_to 'Begin draw process', activate_draw_path(@draw), method: :patch %>
+  <%= link_to 'Begin draw process<br>'.html_safe, activate_draw_path(@draw), method: :patch %>
+<% end %>
+<% if policy(@draw).update? %>
```
orenyk: We're going to want to replace this with `policy(@draw).group_actions?` after #165 is merged in.

orenyk: Also, wouldn't this not apply to students at all?

app/views/draws/show.html.erb
```diff
@@ -19,7 +19,10 @@
...
+  <%= link_to 'Add group to draw<br>'.html_safe, new_draw_group_path(@draw) %>
```
orenyk: Why are we adding line breaks to the link text? We should either do it outside or simply style these links with CSS to have some more padding.

app/views/suites/show.html.erb
```diff
@@ -1,3 +1,10 @@
 <h1 class="suite-number"><%= @suite.number %></h1>
+<h2>Rooms</h2>
+<ul>
+  <% @rooms.each do |room| %>
+    <li><%= link_to room.number, room_path(room) %> - <%= room.beds %> beds</li>
```
orenyk: Same comment re: pluralize.

app/views/buildings/show.html.erb
esoterik: @spaceraccoon there's a built-in function for this, just so you know

spaceraccoon:
@esoterik @orenyk Fixed Oren's comments but travis/rake keeps giving me 'must be logged in' fails

spaceraccoon:
@orenyk @esoterik All checks passed and ready for re-review.

app/views/application/_nav.html.erb
```diff
@@ -7,6 +7,31 @@
...
+        <ul class="dropdown menu" data-dropdown-menu>
+          <% if current_user && policy(Draw).index? %>
```
orenyk: Ok, since all of these links depend on `current_user` we should extract that conditional to before the `ul.dropdown.menu` element.

Then, in the interest of time, we're going to build two menus, one for admins and one for students, and extract them as partials. We're going to put an example of this in the branch.

PR #160, opened 2017-02-08
[154] Implement admin editing of drawless groups

### LINKED ISSUE #154: Allow admins to edit drawless groups (or groups in general?)
There is currently no link 😞 

Also, after further testing, I tried to add a user to the drawless group and it threw an `ActiveRecord::RecordInvalid` exception on the membership (both because of a draw mismatch and bad user intent). This opens up two issues:

1. Drawless group updating should probably have a separate service object rather than Updater since we'll likely want to allow for a) user draw deletion and b) maybe even user intent switching to `on_campus` if it is currently `undeclared` (this last point may be true during creation as well).
2. More importantly, it seems like our has many through association calls `save!` when creating memberships, which prevents us from catching these errors naturally. We need to figure this out.

orenyk:
Resolves #154
- Add DrawlessGroupUpdater service object
- Add old_draw_id parameter to users

~~Still needs the DrawlessGroupCreator updated to store existing student draws in `old_draw_id`, but otherwise this should be pretty close.~~ I think this is good to go!

### DIFF
--- a/app/services/creators/drawless_group_creator.rb
+++ b/app/services/creators/drawless_group_creator.rb
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 #
 # Service object for 'special' (non-draw) housing groups
+# TODO: If we end up adding more shared behavior between GroupCreator and this
+# class, don't duplicate and either use inheritance or a module to keep things
+# DRY.
 class DrawlessGroupCreator < Creator
   # Initialize a new SpecialGroupCreator
   #
@@ -32,17 +35,22 @@ class DrawlessGroupCreator < Creator
   def process_params
     @params = params.to_h.transform_keys(&:to_sym)
     remove_blank_members if params[:member_ids]
+    remove_remove_ids_from_params if params[:remove_ids]
   end
 
-  # TODO: If we end up adding more shared behavior between GroupCreator and this
-  # class, don't duplicate and either use inheritance or a module to keep things
-  # DRY.
   def remove_blank_members
     @params[:member_ids] = params[:member_ids].reject(&:empty?)
   end
 
+  def remove_remove_ids_from_params
+    @params.delete(:remove_ids)
+  end
+
+  # Note that this occurs within the transaction
   def ensure_valid_members
-    User.where(id: all_member_ids).update_all(draw_id: nil, intent: 'on_campus')
+    User.where(id: all_member_ids).each do |user|
+      user.remove_draw.update!(intent: 'on_campus')
+    end
   end
 
   def all_member_ids
--- a/app/services/creators/group_creator.rb
+++ b/app/services/creators/group_creator.rb
@@ -16,6 +16,7 @@ class GroupCreator < Creator
     @params = params.to_h.transform_keys(&:to_sym)
     add_draw_to_params if params[:leader_id]
     remove_blank_members if params[:member_ids]
+    remove_remove_ids_from_params if params[:remove_ids]
   end
 
   def add_draw_to_params
@@ -24,6 +25,10 @@ class GroupCreator < Creator
 
   def remove_blank_members
     @params[:member_ids] = params[:member_ids].reject(&:empty?)
+  end
+
+  def remove_remove_ids_from_params
+    @params.delete(:remove_ids)
   end
 
   def success(group)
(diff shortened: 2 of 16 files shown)


orenyk:
@esoterik ready for review!

app/services/creators/drawless_group_creator.rb
```diff
@@ -41,8 +42,16 @@ def remove_blank_members
     @params[:member_ids] = params[:member_ids].reject(&:empty?)
   end
 
+  # TODO: does this count or should we still leave it?
+  def remove_remove_ids_from_params
```
esoterik: this doesn't need to be a separate method

app/services/creators/drawless_group_creator.rb
```diff
@@ -41,8 +42,16 @@ def remove_blank_members
...
+    @params.delete(:remove_ids)
+  end
+
   def ensure_valid_members
```
esoterik: can you document that this is meant to be run in the transaction?

app/services/creators/group_creator.rb
```diff
@@ -26,6 +27,10 @@ def remove_blank_members
     @params[:member_ids] = params[:member_ids].reject(&:empty?)
   end
 
+  def remove_remove_ids_from_params
```
esoterik: this also doesn't need to be a separate method

app/services/updaters/drawless_group_updater.rb
```diff
@@ -0,0 +1,88 @@
...
+
+  def delete_leader_id_if_empty
+    params.delete(:leader_id) if params[:leader_id]
+  end
+
+  def remove_users
+    Membership.where(user_id: pending_users[:remove].map(&:id)).destroy_all
+    pending_users[:remove].each do |user|
+      old_draw_id = user.old_draw_id
+      user.update!(draw_id: old_draw_id, old_draw_id: nil, intent: 'undeclared')
```
esoterik: should the body of this block be a method on the user model?

app/services/updaters/drawless_group_updater.rb
```diff
@@ -0,0 +1,88 @@
...
+    end
+  end
+
+  def add_users
+    pending_users[:add].each do |user|
+      draw_id = user.draw_id
+      user.update!(draw_id: nil, old_draw_id: draw_id, intent: 'on_campus')
```
esoterik: same with this block

spec/services/creators/drawless_group_creator_spec.rb
```diff
@@ -14,6 +14,14 @@
       expect(described_class.new(params).create![:object]).to \
         be_instance_of(Group)
     end
+    it 'assigns the :old_draw_id of the students to their original draws' do
+      params = instance_spy('ActionController::Parameters',
+                            to_h: params_hash_with_draw_students)
+      leader = described_class.new(params).create![:object].leader
+      # rubocop:disable RSpec/InstanceVariable
+      expect(leader.old_draw_id).to eq(@leader.draw_id)
```
esoterik: document why this is an instance variable

spec/services/updaters/drawless_group_updater_spec.rb
```diff
@@ -0,0 +1,103 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.describe DrawlessGroupUpdater do
+  # Delete memberships of removed members before attempting to update
+  # Remove added students to draws if they are already in one
+  # switch removed students back to original draw
+  #   add old_draw_id attribute to users
+  #   update drawless_groups_creator to assign that value
+  # update intent of added students to on_campus
```
esoterik: do we still need this comment?

spec/services/updaters/drawless_group_updater_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+      it 'sets an error message' do
+        group = instance_spy('group')
+        allow(group).to receive(:update!).and_raise(ActiveRecord::RecordInvalid)
+        p = instance_spy('ActionController::Parameters', to_h: { size: 4 })
+        result = described_class.update(group: group, params: p)
+        expect(result[:msg]).to have_key(:error)
+      end
+    end
+
+    def valid_params
```
esoterik: you never filled out this method

PR #163, opened 2017-02-09
[159] Prevent admins from removing a group leader

### LINKED ISSUE #159: Prevent Leaders from leaving/being removed groups
A new leader should be declared before the current leader can leave

### LINKED ISSUE #161: Fix Group#members to only pull from accepted memberships
We started this in #141 and took care of `User#membership` but did not update the Group model.

orenyk:
Resolves #159
- Add Group#removable_members helper method
- Add :accepted_memberships conditional has_many association to use for has_many :members, through: :accepted_memberships
- Add @leader_students variable to GroupsController and DrawlessGroipsController for leader dropdown

Blocked on merging of #160, will need to be rebased following that merge.

Also resolves #161

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -12,7 +12,9 @@ class Group < ApplicationRecord
   belongs_to :draw
   has_one :suite
   has_many :memberships, dependent: :delete_all
-  has_many :members, through: :memberships, source: :user
+  has_many :full_memberships, -> { where(status: 'accepted') },
+           class_name: 'Membership', inverse_of: :group
+  has_many :members, through: :full_memberships, source: :user
 
   enum status: %w(open full locked)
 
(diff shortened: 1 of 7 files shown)


app/models/group.rb
```diff
@@ -12,7 +12,9 @@ class Group < ApplicationRecord
   belongs_to :draw
   has_one :suite
   has_many :memberships, dependent: :delete_all
-  has_many :members, through: :memberships, source: :user
+  has_many :full_memberships, -> { where(status: 'accepted') },
+           class_name: 'Membership', inverse_of: :group
```
esoterik: does this also need a `dependent: :delete_all`?

esoterik: does this also need a `dependent: :delete_all`?

orenyk: This is tested and still passed, so I'm going to say it's not necessary: https://github.com/YaleSTC/vesta/blob/159_leaders_are_safe/spec/models/group_spec.rb#L65

Feel free to merge if there was nothing else!

orenyk:
@esoterik ⬆️

esoterik:
merging!

PR #165, opened 2017-02-09
[120] Prevent group actions on non-pre-lottery draws

### LINKED ISSUE #120: Prevent actions on draft draws
Certain actions aren't permitted on draft draws, like Group creation. We should figure out if there are other currently permitted actions and update the policies as needed.

orenyk:
Resolves #120
- Add after(:create) callback to :open_group and :full_group factories to automatically update the draw status to pre_lottery
- Add :group_actions? permission to DrawPolicy and call it in the authorize! method of GroupsController


esoterik:
rebased and ready to go

PR #192, opened 2017-02-13
[175] Add suite summary page to draws

orenyk:
Resolves #175
- Add suite summary page with a listing of all of the suites belonging
  to the draw, separated by suite size
- Add SuitesBySize query object

PR #196, opened 2017-02-14
[191] Add bulk suite-editing for draws

### LINKED ISSUE #191: Admins should be able to bulk-update the suites in a draw
Extension of #175

orenyk:
Resolves #191
- add draws_suites_edit and draws_suites_update routes / actions
- Add DrawSuitesUpdate form / service object
- Add UndrawnSuitesQuery
- Add DrawnSuitesForDrawQuery

Blocked on #192

### DIFF
--- a/app/models/draw_suites_update.rb
+++ b/app/models/draw_suites_update.rb
@@ -1,0 +1,108 @@
+# frozen_string_literal: true
+#
+# Form / service object to handle the updating of draw suites
+class DrawSuitesUpdate
+  include ActiveModel::Model
+
+  attr_reader :size, :suite_ids, :drawn_suite_ids, :undrawn_suite_ids
+
+  # permit :update to be called on the class object
+  def self.update(**params)
+    new(**params).update
+  end
+
+  # Initialize a new DrawSuitesUpdate
+  #
+  # @param draw [Draw] the draw to be updated
+  # @param params [#to_h] the parameters from the form
+  def initialize(draw:, params: nil)
+    @draw = draw
+    @suite_ids = draw.suite_ids
+    process_params(params) if params
+  end
+
+  # Execute the suites update, remove all suites to be removed and add all
+  # suites to be added. Occurs in a transaction for safety.
+  #
+  # @return [Hash{Symbol=>String,Hash,Nil,DrawStringUpdate] a result hash
+  #   containing the appropriate path to redirect to, a flash message to set,
+  #   and the DrawSuitesUpdate object if there were any failures.
+  def update
+    return no_action_warning if suites_to_remove.empty? && suites_to_add.empty?
+    ActiveRecord::Base.transaction do
+      remove_suites unless suites_to_remove.empty?
+      add_suites unless suites_to_add.empty?
+    end
+    success
+  rescue ActiveRecord::RecordInvalid => error
+    error(error)
+  end
+
+  private
+
+  attr_reader :draw, :params, :suites_to_add, :suites_to_remove
+
+  def process_params(params)
+    @params = params.to_h.transform_keys(&:to_sym)
+    @size = extract_size
+    update_ids_param(:suite_ids)
+    update_ids_param(:drawn_suite_ids)
+    update_ids_param(:undrawn_suite_ids)
+    @suites_to_remove = find_suites_to_remove
+    @suites_to_add = find_suites_to_add
+  end
+
+  def extract_size
+    raise ArgumentError if size_param_not_an_int
+    params[:size].to_i
+  end
+
+  def size_param_not_an_int
+    params[:size].to_s.match(/\d+/).nil?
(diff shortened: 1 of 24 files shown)


app/queries/undrawn_suites_query.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+#
+# Query to return the suites that don't belong to any draw. This can bee passed
+# an existing relation for a subset of all Suites.
+class UndrawnSuitesQuery
```
esoterik: I don't know if I like the drawn/undrawn naming -- I think drawless might be more appropriate here?

spec/features/draws/draw_suite_update_spec.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.feature 'Draw suite update' do
+  let(:draw) { FactoryGirl.create(:draw_with_members, suites_count: 1) }
+  let!(:removed_suite) { draw.suites.first }
+  let(:other_draw) { FactoryGirl.create(:draw_with_members, suites_count: 1) }
+  let!(:other_draw_suite) { other_draw.suites.first }
```
orenyk: This is because without the bang it only executes when called, but the bang executes prior to the tests. Since `other_draw_suite` refers to `other_draw`, by applying the bang there it effectively causes both to execute.

This was necessary since the views require the data to be persisted prior to rendering and we only refer to the variables in the assertion.

app/queries/undrawn_suites_query.rb
orenyk: Sure, I can update that.

orenyk: Do you want to update `drawn`? I don't think `drawed` makes sense.

orenyk:
Updated the service object to only remove available suites (with a test).

app/models/draw_suites_update.rb
```diff
@@ -0,0 +1,108 @@
...
+
+  def update_ids_param(key)
+    params[key] = [] unless params[key]
+    @params[key] = params[key].reject(&:empty?)
+  end
+
+  def find_suites_to_remove
+    return [] unless params[:suite_ids]
+    # TODO: refactor not to require an extra db hit here, also makes testing
+    # better
```
orenyk: I'll just do it now so we don't forget

app/models/draw_suites_update.rb
```diff
@@ -0,0 +1,108 @@
+# frozen_string_literal: true
+#
+# Form / service object to handle the updating of draw suites
+class DrawSuitesUpdate
+  include ActiveModel::Model
+
+  attr_reader :size, :suite_ids, :drawn_suite_ids, :undrawn_suite_ids
+
+  # permit :update to be called on the parent class
```
esoterik: nitpicky but it's not a parent class it's an instance of the class

app/queries/drawn_suites_for_draw_query.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+#
+# Query to return the suites that belong to at least one draw, with the
+# exclusion of those that belong to the passed draw.
+class DrawnSuitesForDrawQuery
```
esoterik: maybe `SuitesInOtherDrawsQuery` is a better name?

app/models/draw_suites_update.rb
```diff
@@ -0,0 +1,108 @@
...
+    @size = extract_size
+    update_ids_param(:suite_ids)
+    update_ids_param(:drawn_suite_ids)
+    update_ids_param(:undrawn_suite_ids)
+    @suites_to_remove = find_suites_to_remove
+    @suites_to_add = find_suites_to_add
+  end
+
+  def extract_size
+    raise ArgumentError if size_param_not_an_int
```
esoterik: I don't like raising an exception here, but I'll open another issue for it

app/models/draw_suites_update.rb
orenyk: It's not an instance, that would be `def update`, but it's also not the parent class. I'll update it to `class object`.

app/models/draw_suites_update.rb
orenyk: I mean, if we don't raise here I'm pretty sure that the queries that make use of the size will raise errors anyway since ActiveRecord won't be able to cast the param to an integer. We can talk more about this after pilot.

app/queries/drawn_suites_for_draw_query.rb
orenyk: Sure, I'll update.

esoterik:
waiting on green build to merge!

esoterik:
merging

PR #197, opened 2017-02-14
[158] Restore student draws when destroying drawless groups

### LINKED ISSUE #158: Return members of Drawless Groups to their Draws on group deletion

orenyk:
Resolves #158 
- Add after_destroy callback on Group for drawless groups that restores the draws for each of its members

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -31,6 +31,8 @@ class Group < ApplicationRecord
 
   before_validation :add_leader_to_members, if: ->(g) { g.leader.present? }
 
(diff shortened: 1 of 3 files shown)


esoterik:
@orenyk did we open a PR / otherwise document the intermittent failure in that occurred here?

orenyk:
I did not - it appears to be a race condition. I feel like we could just do `.map(&:id).sort` to get consistent results if you'd prefer.

esoterik:
ok, I'll make that change here and merge this in

esoterik:
ok rebased and fixed the race condition, waiting on green build to squash and merge

PR #198, opened 2017-02-14
[118] Write spec to ensure group creation isn't broken

### LINKED ISSUE #118: Prevent creation of groups with too many members
Steps to Recreate:

```ruby
> group = Group.new(size: 1, leader_id; id1, member_ids: [id2])
```

Expected:

```ruby
> group.save #=> false
> group.valid? #=> false
```

Actual:

```ruby
> group.save #=> false
> group.persisted? #=> true
> group.valid? #=> false
```

I'm pretty sure this has to do with the counter cache on memberships not being updated until _after_ the group is created (since it defaults to 0). Pretty important...

orenyk:
Resolves #118 

This apparently is no longer broken? I think the spec is worth keeping, even though it broke at one point while coding but now seems to be passing.

### DIFF
--- a/spec/models/group_spec.rb
+++ b/spec/models/group_spec.rb
@@ -42,6 +42,13 @@ RSpec.describe Group, type: :model do
       group.size = 1
       expect(group.valid?).to be_falsey
     end
+    it 'also validates during creation' do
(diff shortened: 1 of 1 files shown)


esoterik:
interesting! merging

PR #218, opened 2017-02-21
[179] Add suite merging functionality

### LINKED ISSUE #179: Admins can merge suites
Use case: grouping many standalone singles into a suite, adding a standalone single to an existing suite as a bedroom.

orenyk:
Resolves #179
- Add SuiteMergerForm form / service object
- Add OtherSuitesInBuildingQuery query object
- Add routes / view / policies for suite merger
- Update suite size counter cache callbacks in Room to handle switching the suite

### DIFF
--- a/app/forms/suite_merger_form.rb
+++ b/app/forms/suite_merger_form.rb
@@ -1,0 +1,142 @@
+# frozen_string_literal: true
+#
+# Form / service object for merging two suites together. Ensures suites belong
+# to the same building.
+class SuiteMergerForm
+  include ActiveModel::Model
+
+  attr_reader :other_suite_number, :other_suite
+  attr_writer :number
+
+  validates :suite, presence: true
+  validates :other_suite, presence: { message: 'must be a valid suite' }
+  validates :number, presence: true
+  validates :other_suite_number, presence: true
+  validate :both_suites_different
+  validate :both_suites_in_same_building
+  validate :both_suites_available
+
+  # Allows for calling :submit on the class object
+  def self.submit(**params)
+    new(**params).submit
+  end
+
+  # Initialize a new SuiteMergerForm
+  # @param suite [Suite] the base suite to merge into
+  # @param params [#to_h] the form params from the controller
+  def initialize(suite:, params: nil)
+    @suite = suite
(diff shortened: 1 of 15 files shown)


spec/forms/suite_merger_form_spec.rb
```diff
@@ -0,0 +1,135 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.describe SuiteMergerForm, type: :model do
+  describe '.submit' do
+    xit 'calls #submit on a new instance of SuiteMergerForm'
+  end
+
+  describe 'validations' do
+    subject { described_class.new }
```
orenyk: delete this

app/forms/suite_merger_form.rb
```diff
@@ -0,0 +1,135 @@
+# frozen_string_literal: true
+#
+# Form / service object for merging two suites together. Ensures suites belong
+# to the same building.
+class SuiteMergerForm
+  include ActiveModel::Model
+
+  attr_reader :other_suite_id
+  attr_writer :number
+  attr_accessor :suite, :other_suite
```
esoterik: do these need to be public?

app/views/suites/show.html.erb
```diff
@@ -6,5 +6,11 @@
   <% end %>
 </ul>
 <p><%= link_to 'Add room', new_room_path %></p>
+<% if policy(@suite).merge? %>
```
esoterik: can we add some sort of header / instructions to this?

orenyk:
Ok, this should be fixed to take a field and hopefully exhibit somewhat sane behavior when it comes to input issues. Currently the form on the suite show page will take you to the full form _even if the suite number entered is invalid_. This seemed acceptable to me since the final merge will fail if it isn't found with a reasonable error message. What do you think?

orenyk:
Updated!

app/views/suites/merge.html.erb
```diff
@@ -0,0 +1,12 @@
+<h1>Merge suite <%= @suite.number %></h1>
```
esoterik: Can the header have both suite numbers in it?

PR #227, opened 2017-02-22
[219] Fix Draw#bed_count to include suite availability

### LINKED ISSUE #219: Draw#beds_count doesn't take suite availability into account
Currently, `Draw#beds_count` sums over all suites in the draw, but it should really only sum over suites without groups assigned (or possibly suites without groups assigned externally to the draw).

orenyk:
Resolves #219 

Since `bed_count` is a private method it wasn't currently tested, and we were stubbing it out to test `enough_beds?`, there were no tests for this. That smells a bit to me, let me know if you want to either make `bed_count` public or test it somehow;

### DIFF
--- a/app/models/draw.rb
+++ b/app/models/draw.rb
@@ -48,7 +48,10 @@ class Draw < ApplicationRecord
     @student_count ||= students.count
   end
 
+  # Calculate the number of beds that exist across all available suites
+  #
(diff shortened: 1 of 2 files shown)


esoterik:
adding a test would be good!

orenyk:
Even though it's private? Should we just make it public?

orenyk:
Added a test!

PR #231, opened 2017-02-23
[57] [WIP] Group locking

esoterik:
Two small, mostly front-end things to be done still:

- [x] Add a finalize membership button that shows conditionally
- [x] Add an admin / housing rep only button to lock any group as long as it is full, with a test

But, I think this is ready for a review!

### DIFF
--- a/app/models/draw.rb
+++ b/app/models/draw.rb
@@ -34,6 +34,13 @@ class Draw < ApplicationRecord
     suite_sizes - locked_sizes
   end
 
+  # Query method get the suites without groups in the draw
+  #
+  # @return [ActiveRecord::Relation] the suites without assigned groups
+  def available_suites
+    suites.available
+  end
+
   # Query method to see if a draw has at least one student.
   #
   # @return [Boolean] whether or not the draw has at least one student
--- a/app/services/group_finalizer.rb
+++ b/app/services/group_finalizer.rb
@@ -1,0 +1,65 @@
+# frozen_string_literal: true
+#
+# Service object to finalize groups
+class GroupFinalizer
+  # Initialize a new GroupFinalizer and call #finalize on it
+  def self.finalize(**params)
+    new(**params).finalize
+  end
+
+  # Initialize a GroupFinalizer
+  #
+  # @param group [Group] The group object to be updated
+  def initialize(group:)
+    @group = group
+    @errors = []
+  end
+
+  # Finalize a group
+  #
+  # @return [Hash{Symbol=>Array, Group, Hash}] The result of the finalizing
+  def finalize # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+    return error unless valid?
+    ActiveRecord::Base.transaction do
+      group.finalizing!
+      unless group.leader.membership.locked
+        group.leader.membership.update!(locked: true)
+      end
+    end
+    success
+  rescue ActiveRecord::RecordInvalid => failures
+    @errors = failures
+    error
+  end
+
+  private
+
+  attr_accessor :errors
+  attr_reader :group
+
+  def valid?
+    group_full?
+    suite_size_open?
+    errors.empty?
+  end
+
+  def group_full?
+    return if group.full?
+    @errors << 'Group must be full'
+  end
+
+  def suite_size_open?
+    return if group.draw.open_suite_sizes.include? group.size
+    @errors << 'Suite size must be open'
+  end
+
+  def error
+    { object: [group.draw, group], record: group,
+      msg: { error: errors.join(', ') } }
+  end
+
+  def success
+    { object: [group.draw, group], record: group,
+      msg: { success: "#{group.name} is being finalized." } }
+  end
+end
--- a/spec/features/groups/group_freezing_spec.rb
+++ b/spec/features/groups/group_freezing_spec.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.feature 'Group locking' do
+  context 'suite size still available, full group' do
+    it 'can be initiated by a leader' do
+      group = FactoryGirl.create(:full_group, size: 2)
+      log_in group.leader
+      visit draw_group_path(group.draw, group)
+      click_on 'Finalize Group'
+      expect(locks_leader_and_finalizes_group(group: group)).to be_truthy
+    end
+    it 'can be confirmed by a member' do
+      group = FactoryGirl.create(:finalizing_group, size: 2)
+      log_in group.members.last
+      visit draw_group_path(group.draw, group)
+      click_on 'Finalize Membership'
+      expect(group.reload).to be_locked
+    end
+
+    def locks_leader_and_finalizes_group(group:)
+      group.reload
+      (group.locked_members - [group.leader]).empty? && group.finalizing?
+    end
+  end
+  context 'as admin, full group' do
+    it 'can be locked' do
+      group = FactoryGirl.create(:full_group, size: 2)
+      log_in FactoryGirl.create(:admin)
+      visit draw_group_path(group.draw, group)
+      click_on 'Lock Group'
+      expect(group.reload).to be_locked
+    end
+  end
+end
--- a/spec/models/draw_spec.rb
+++ b/spec/models/draw_spec.rb
@@ -78,4 +78,13 @@ RSpec.describe Draw, type: :model do
       expect(draw.send(:bed_count)).to eq(draw.suites.first.size)
     end
   end
+
+  describe '#available_suites' do
+    it 'returns suites without associated groups' do
+      available = FactoryGirl.create(:suite)
+      taken = FactoryGirl.create(:locked_group, :with_suite).suite
+      draw = FactoryGirl.create(:draw, suites: [available, taken])
+      expect(draw.available_suites).to eq([available])
+    end
+  end
 end
--- a/spec/models/group_spec.rb
+++ b/spec/models/group_spec.rb
@@ -57,6 +57,12 @@ RSpec.describe Group, type: :model do
       group.status = 'locked'
       expect(group.valid?).to be_falsey
     end
+    it 'can only be locked if all memberships are locked' do
+      # finalizing locks the leader only, not any of the members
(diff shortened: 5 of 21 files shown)


spec/features/groups/group_freezing_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+      member = group.members.last
+      log_in member
+      visit draw_group_path(group.draw, group)
+      click_on 'Finalize Membership'
+      expect(group.locked_members).to eq(group.members)
+      expect(group.locked?).to be_truthy
+    end
+  end
+  context 'suite size not available, as group leader' do 
+    it 'cannot be initiated' do
```
esoterik: I think I decided against testing this failure case, but let me know if you want this text

app/models/draw.rb
```diff
@@ -27,6 +27,17 @@ def open_suite_sizes
...
+  # Query method get the suites without groups in the draw
+  #
+  # @return [ActiveRecord::Relation] the suites without assigned groups
+  def available_suites
+    suites.available
+  end
+
+  def available_suite_sizes
+    available_suites.map(&:size).uniq
+  end
```
esoterik: this should all be deleted -- forgot to do it after I rebased

app/services/group_finalizer.rb
```diff
@@ -0,0 +1,59 @@
...
+    @errors = failures
+    error
+  end
+
+  private
+
+  attr_reader :group, :errors
+
+  def valid?
+    group_full
```
orenyk: Hmm it occurs to me that if you ran `valid?` multiple times that it would just keep adding stuff to the errors array. Not sure if there's a way around this (and I did something similar in `DrawActivator`, but that doesn't mean it's good. Definitely fine for now, but we should open an issue to fix this (maybe call it in the initializet?).

app/services/group_finalizer.rb
```diff
@@ -0,0 +1,59 @@
...
+    suite_size_available
+    return false unless errors.empty?
+    true
+  end
+
+  def group_full
+    return if group.full?
+    @errors += ['Group must be full']
```
orenyk: Why wouldn't you use an `attr_accessor` for errors? Also, why `+=` instead of `<<`?

app/services/group_finalizer.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def suite_size_available
+    return if group.draw.available_suite_sizes.include? group.size
```
orenyk: Should we be using `open_suite_sizes` here instead?

app/services/group_finalizer.rb
```diff
@@ -0,0 +1,59 @@
...
+  attr_reader :group, :errors
+
+  def valid?
+    group_full
+    suite_size_available
+    return false unless errors.empty?
+    true
+  end
+
+  def group_full
```
orenyk: Shouldn't this be `group_full?`? I know it's a minor issue but these feel like query methods to me.

app/services/group_finalizer.rb
```diff
@@ -0,0 +1,59 @@
...
+    return if group.full?
+    @errors += ['Group must be full']
+  end
+
+  def suite_size_available
```
orenyk: Same here (`suite_size_available?`)

app/services/group_finalizer.rb
```diff
@@ -0,0 +1,59 @@
...
+    return if group.draw.available_suite_sizes.include? group.size
+    @errors += ['Suite size must be available']
+  end
+
+  def error
+    { object: [group.draw, group], record: group,
+      msg: { error: errors.join(' ') } }
```
orenyk: Shouldn't we be joining with commas here? At least, isn't that what we've been doing?

spec/features/groups/group_freezing_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+require 'rails_helper'
+
+RSpec.feature 'Group locking' do
+  context 'suite size still available, full group' do
+    it 'can be initiated by a leader' do
+      group = FactoryGirl.create(:full_group, size: 2)
+      log_in group.leader
+      visit draw_group_path(group.draw, group)
+      click_on 'Finalize Group'
+      expect(group.locked_members).to eq([group.leader])
```
orenyk: Isn't this dual expectation going to fail rubocop? Maybe turn it into separate method?

spec/features/groups/group_freezing_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    end
+    it 'can be confirmed by a member' do
+      group = FactoryGirl.create(:full_group, size: 2)
+      GroupFinalizer.finalize(group: group)
+      member = group.members.last
+      log_in member
+      visit draw_group_path(group.draw, group)
+      click_on 'Finalize Membership'
+      expect(group.locked_members).to eq(group.members)
+      expect(group.locked?).to be_truthy
```
orenyk: `expect(group).to be_locked`?

orenyk: Also, does this need to be `group.reload`?

spec/features/groups/group_freezing_spec.rb
orenyk: I don't think it needs to be tested in a feature spec.

spec/models/draw_spec.rb
```diff
@@ -69,4 +69,22 @@
       expect(draw.open_suite_sizes).to eq([2])
     end
   end
+
+  describe '#available_suites' do
```
orenyk: Don't forget to remove these if you take out the methods (although RSpec should pick up on it either way)

spec/models/group_spec.rb
```diff
@@ -57,6 +57,11 @@
       group.status = 'locked'
       expect(group.valid?).to be_falsey
     end
+    it 'can only be locked if all memberships are locked' do
+      group = FactoryGirl.create(:finalizing_group)
```
orenyk: The connection between this setup and the test conditions is unclear, probably because it's all included in the factory. Can we be more explicit?

spec/models/membership_spec.rb
```diff
@@ -132,12 +138,26 @@
...
+      group = FactoryGirl.create(:finalizing_group)
+      membership = group.memberships.first
+      expect { membership.destroy }.not_to change { group.memberships_count }
+    end
+    it 'must be accepted' do
+      group = FactoryGirl.create(:finalizing_group)
+      membership = Membership.new(group: group,
+                                  user: FactoryGirl.create(:student,
+                                                           draw: group.draw))
+      membership.update(status: 'requested')
```
orenyk: Can't this be part of the FactoryGirl call?

spec/models/membership_spec.rb
```diff
@@ -132,12 +138,26 @@
...
+      expect(membership.valid?).to be_falsey
+    end
+    it 'must belong to a finalizing group' do
       group = FactoryGirl.create(:full_group)
-      group.update_attributes(status: 'locked')
-      expect { group.memberships.first.destroy }.not_to \
-        change { group.memberships_count }
+      membership = group.memberships.first
+      membership.locked = true
+      expect(membership.valid?).to be_falsey
```
orenyk: `expect(membership).not_to be_valid`? (also above)

spec/services/group_finalizer_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+      described_class.finalize(group: group)
+      expect(group).to have_received(:finalizing!)
+    end
+    it 'returns the redirect objects in :object' do
+      results = described_class.finalize(group: group)
+      expect(results[:object]).to eq([group.draw, group])
+    end
+    it 'returns a success msg' do
+      results = described_class.finalize(group: group)
+      expect(results[:msg].keys).to eq([:success])
```
orenyk: I think we've been using `expect(results[:msg]).to have_key(:success)`

esoterik: I know but I think this is a stronger assertion

spec/models/membership_spec.rb
```diff
@@ -132,12 +138,26 @@
...
+  context 'locked membership' do
     it 'cannot be destroyed' do
+      group = FactoryGirl.create(:finalizing_group)
+      membership = group.memberships.first
+      expect { membership.destroy }.not_to change { group.memberships_count }
+    end
+    it 'must be accepted' do # rubocop:disable RSpec/ExampleLength
+      group = FactoryGirl.create(:finalizing_group)
+      membership = Membership.new(group: group,
+                                  user: FactoryGirl.create(:student,
```
AlanLiu96: very very minor thing, but just to be consistent, I would probably define user separately like group was defined. Then the line spacing isn't as awkward

spec/models/membership_spec.rb
```diff
@@ -132,12 +138,26 @@
...
+                                                           draw: group.draw),
+                                  status: 'requested')
+      membership.locked = true
+      expect(membership).not_to be_valid
+    end
```
AlanLiu96: # rubocop:enable RSpec/ExampleLength

orenyk: You don't need to re-enable unless it was on it's own line - the way it's written above it only suppresses the error on the that specific line.

spec/models/membership_spec.rb
orenyk: Let's fix this.

spec/services/group_finalizer_spec.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.describe GroupFinalizer do
+  describe '.finalize' do
+    xit 'calls #finalize on a new instance' do
```
orenyk: You don't need to pass a block for a pending test.

spec/services/group_locker_spec.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.describe GroupLocker do
+  describe '.lock' do
+    xit 'calls #lock on a new instance' do
```
orenyk: No block necessary

PR #238, opened 2017-02-26
[20] Adds Suite CSV Import

### LINKED ISSUE #20: Admins can Bulk-add Facilities Information
We need to figure out a good format for this. We might be using some internal API for usage at Yale, but we should implement something more general as well.

Blocked on #9.

esoterik:
Resolves #20
  - Admins can upload a CSV of suite information to import room data
  - Adds SuiteImporter form object


app/models/suite_importer.rb
```diff
@@ -0,0 +1,132 @@
...
+  end
+
+  # Read in a CSV of suite information and create suites and rooms accordingly
+  # The CSV should be formatted with a header Number, Single, Double, where
+  # Single is the number of single rooms in the suite and Double is the number
+  # of double rooms.
+  #
+  # @return [Hash{Symbol=>nil,Hash}] A hash with flash messages to be set.
+  def import
+    return error('No file uploaded') unless file
```
orenyk: so... ideally we would be using ActiveModel validations for this stuff, e.g. `validate :file, presence: true` and `validates :correct_header?` (probably renamed without the question mark). Then we'd just return `error(errors.full_messages.join(', ')) unless valid?` at the start of `#submit`.

app/models/suite_importer.rb
```diff
@@ -0,0 +1,132 @@
...
+      create_rooms(suite: suite, singles: row['singles'].to_i,
+                   doubles: row['doubles'].to_i)
+    end
+    successes << row_number
+  rescue
+    failures << row_number
+  end
+
+  def create_rooms(suite:, doubles:, singles:)
+    total = doubles + singles
```
orenyk: Add a comment for why counting down

app/views/buildings/show.html.erb
```diff
@@ -5,5 +5,17 @@
     <li><%= link_to suite.number, suite_path(suite) %> - <%= pluralize(suite.size,'bed') %></li>
   <% end %>
 </ul>
+<h2>Import Suites</h2>
+<p>Add many suites at once by uploading a CSV file with the building information.</p>
+<p>The CSV must have the format "Number, Singles, Doubles" where Number is the
```
orenyk: Make this just one `<p>`

app/views/buildings/show.html.erb
```diff
@@ -5,5 +5,17 @@
     <li><%= link_to suite.number, suite_path(suite) %> - <%= pluralize(suite.size,'bed') %></li>
   <% end %>
 </ul>
+<h2>Import Suites</h2>
+<p>Add many suites at once by uploading a CSV file with the building information.</p>
+<p>The CSV must have the format "Number, Singles, Doubles" where Number is the 
+identifier of the suite, Singles is the number of single bedrooms in the suite,
+and Doubles is the number of double bedrooms in the suite.</p>
+<p>
+<%= simple_form_for @suite_importer, url: suite_import_path(@suite_importer) do |f| %>
```
orenyk: Why are you passing `@suite_importer` to `suite_import_path`? I don't think it needs any parameters to generate the route.

spec/models/suite_importer_spec.rb
```diff
@@ -0,0 +1,92 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.describe SuiteImporter do
+  include ActionDispatch::TestProcess
+  describe '.import' do
+    xit 'calls #import on a new instance' do
```
orenyk: Remove block

spec/models/suite_importer_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+    end
+  end
+
+  context 'valid csv' do
+    it 'creates the suites' do
+      building = instance_spy('Building')
```
orenyk: this can be a `let` as well as the next line.

spec/models/suite_importer_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+      file = fixture_file_upload(csv_path('suite_upload'), 'text/csv')
+      stub_suite_create
```
orenyk: this should be a before

spec/models/suite_importer_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+    def stub_suite_create
+      allow(Suite).to receive(:create!)
+        .and_return(instance_spy('Suite', number: 'L01'))
+      allow(Room).to receive(:create!)
+    end
+  end
+
+  context 'missing header' do
+    it 'creates no suites' do
+      building = instance_spy('Building')
```
orenyk: same here (extracting `let`s)

PR #244, opened 2017-02-27
[193] Allow admins to add transfer students to groups

### LINKED ISSUE #193: Admin can add transfer student slots to groups
Should be an attribute on group that "saves" room in the group for transfer students and affects the membership count

orenyk:
Resolves #193
- Add transfers attribute to groups
- Add advanced_edit? policy permission for groups
- Replace `memberships_count` references in group validations with `members_count`, which takes transfers into account

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -8,6 +8,9 @@
 # @attr leader [User] The student that represents the Group.
 # @attr members [Array<User>] The members of the group, excluding the leader.
 # @attr draw [Draw] The Draw that this Group is in.
+# @attr memberships_count [Integer] the number of accepted memberships (counter
+#   cache)
+# @attr transfers [Integer] the number of transfer students in the group
 class Group < ApplicationRecord
   belongs_to :leader, class_name: 'User'
   belongs_to :draw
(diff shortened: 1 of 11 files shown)


app/models/group.rb
```diff
@@ -84,11 +93,19 @@ def locked_members
...
+    (members - locked_members).empty? && members_count == size
   end
 
   private
 
+  # override default attribute getter to include transfers
+  def members_count
+    return 0 unless memberships_count || transfers
+    return memberships_count unless transfers
+    return transfers unless memberships_count
```
esoterik: I don't think we need this; `memberships_count` defaults to 0, and I don't think that we can have a group with transfers without a leader and therefore at least one full member.

app/models/group.rb
```diff
@@ -107,7 +124,7 @@ def validate_suite_size_inclusion
   end
 
   def validate_members_count
-    return unless memberships_count > size
+    return unless members_count && members_count > size
```
esoterik: The first condition shouldn't be necessary

app/models/group.rb
orenyk: It was necessary to test shoulda_matchers, although I agree with you that there needs to be at least a leader so we should never have a memberships_count that's nil. I'll update.

app/models/group.rb
orenyk: :+1: this was from before I switched out `members_count`.

orenyk:
Updated!

PR #283, opened 2017-03-03
[139] Add group leader pending membership management

### LINKED ISSUE #139: Group leaders should be able to manage invitations and requests
They should be able to rescind invitations as well as deny requests.

esoterik:
Resolves #139
  - Add MembershipDestroyer

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+# rubocop:disable Metrics/ClassLength
 # Controller for Groups
 class GroupsController < ApplicationController
   layout 'application_with_sidebar'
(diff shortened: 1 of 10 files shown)


app/controllers/groups_controller.rb
```diff
@@ -68,6 +69,13 @@ def accept_invitation
     handle_action(path: draw_group_path(@draw, @group), **result)
   end
 
+  def reject_pending
+    user = User.includes(:membership).find(params['user_id'])
+    membership = user.memberships.where(group: @group).first
```
orenyk: I think we can use `find_by(group: @group)` instead.

app/controllers/groups_controller.rb
```diff
@@ -68,6 +69,13 @@ def accept_invitation
...
+    result = MembershipDestroyer.new(object: membership).destroy
```
orenyk: No class method, e.g. `MembershipDestroyer.destry(object: membership)`?

spec/services/destroyers/membership_destroyer_spec.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.describe MembershipDestroyer do
+  it 'sucessfully destroys a suite' do
```
orenyk: This string needs to be updated.

spec/services/destroyers/membership_destroyer_spec.rb
```diff
@@ -0,0 +1,10 @@
...
+    m = instance_spy('Membership', destroy: true, user: instance_spy('User'))
+    destroyer = described_class.new(object: m)
+    expect(destroyer.destroy[:object]).to be_nil
```
orenyk: Should we also write a test to assert that `destroy` gets called?

orenyk:
Will merge on green build.

PR #285, opened 2017-03-03
[212] update inventory views

esoterik:
Currently broken!!

### DIFF
--- a/app/controllers/buildings_controller.rb
+++ b/app/controllers/buildings_controller.rb
@@ -5,7 +5,7 @@ class BuildingsController < ApplicationController
 
   def show
     @suite_importer = SuiteImportForm.new(building: @building)
-    @suites = @building.suites
+    @suites = @building.suites.order(:size).group_by(&:size)
   end
 
   def new
@@ -14,6 +14,7 @@ class BuildingsController < ApplicationController
 
   def index
     @buildings = Building.all
+    @suites_by_size = @buildings.map { |b| [b.id, b.suites_by_size] }.to_h
   end
 
   def create
--- a/app/views/buildings/_summary_table.html.erb
+++ b/app/views/buildings/_summary_table.html.erb
@@ -1,0 +1,26 @@
+<h3><%= building.name %></h3>
+<table class="building-summary">
+  <thead>
+    <tr>
+      <td>Suite Size</td>
+      <td># Suites</td>
+      <td># Suites Assigned</td>
+      <td># Suites in Draws</td>
+      <td># Suites not in a Draw</td>
+    </tr>
+  </thead>
+  <tbody>
+    <% @suites_by_size[building.id].each do |size, suites| %>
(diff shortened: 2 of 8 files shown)


app/views/buildings/_summary_table.html.erb
```diff
@@ -0,0 +1,26 @@
...
+      <td># Suites in Draws</td>
+      <td># Suites not in a Draw</td>
+    </tr>
+  </thead>
+  <tbody>
+    <% @suites_by_size[building.id].each_value do |size, suites| %>
+      <tr>
+        <td data-role="suite-size"><%= size_str(size).capitalize.pluralize %></td>
+        <td data-role="suite-count"><%= suites.count %></td>
+        <td data-role="assigned-count"><%= suites.where.not(group_id: nil).count %></td>
```
esoterik: this breaks because `suites` is a hash?? no idea what's going on here

app/controllers/buildings_controller.rb
```diff
@@ -5,7 +5,7 @@ class BuildingsController < ApplicationController
 
   def show
     @suite_importer = SuiteImportForm.new(building: @building)
-    @suites = @building.suites
+    @suites = @building.suites.group_by(&:size)
```
orenyk: We want to sort by size, I think.

app/views/buildings/_summary_table.html.erb
orenyk: Suites is an array, not an association (because of the `group_by` call in the association) so we need to use Ruby Enumerable methods.

app/views/buildings/show.html.erb
```diff
@@ -1,10 +1,47 @@
...
 <h2>Suites</h2>
-<ul>
-  <% @suites.each do |suite| %>
-    <li><%= link_to suite.number, suite_path(suite) %> - <%= pluralize(suite.size,'bed') %></li>
-  <% end %>
-</ul>
+<% @suites.keys.each do |size| %>
+  <div class="suites-list suites-<%= size_str(size).pluralize %>">
+    <h3><%= size_str(size).pluralize.titleize %></h3>
+    <table>
```
orenyk: The column widths here are all over the place since it adjusts on a table by table basis. I'll add some styling.

app/views/buildings/_summary_table.html.erb
```diff
@@ -0,0 +1,26 @@
...
+    <tr>
+      <td>Suite Size</td>
+      <td># Suites</td>
+      <td># Suites Assigned</td>
+      <td># Suites in Draws</td>
+      <td># Suites not in a Draw</td>
+    </tr>
+  </thead>
+  <tbody>
+    <% @suites_by_size[building.id].each_value do |size, suites| %>
```
orenyk: `each_value`???? wtf...

orenyk:
will merge when green

PR #293, opened 2017-03-03
[183] Add the ability to start the lottery phase of a draw

### LINKED ISSUE #183: Admins can proceed to lottery phase
Admins should be warned if they are trying to proceed to the lottery phase with oversubscribed suite sizes.

~~Should include things like the total number of suites of each size and the number still available. This is kind of a subset of #104 but at the same time I think there's room for a simpler version on the draw page itself.~~

orenyk:
Resolves #183
- Add DrawLotteryStarter
- Add Suite#selectable?
- Add Draw#groups? and a whole bunch of other query methods

Please make sure there's at least a pending test for all of the new Draw query methods.


esoterik:
@orenyk ready to rebase and merge!

Issue #314, opened 2017-03-04
Admins should be able to set the intent of all students in a draw to On Campus


spaceraccoon:

orenyk:
Blocked on #315 / #339 for policy purposes, those still need to be written.

esoterik:
Going to create a Draw#lottery_or_later? method for policy purposes instead.

PR #339, opened 2017-03-06
[315] Add Admin Intent Lock on Draw

### LINKED ISSUE #315: Admins should be able to prevent all students in a draw from changing their intent

esoterik:
Resolves #315


app/models/draw.rb
```diff
@@ -17,6 +17,8 @@ class Draw < ApplicationRecord
   validates :name, presence: true
   validates :status, presence: true
 
+  validate :cannot_lock_intent_if_undeclared
```
orenyk: So.... this runs all the time, not just when `lock_intent` is being set to true. I'm not sure why the tests didn't pick up on this, but it's causing 500 errors when I try to activate a draw with seeded data.

The reason is because `calculate_oversub_metrics` exits in the original request (since the draw is a draft so `policy(@draw).oversub_report?` is not permitted), but then somehow the change to the draw status persists even though DrawActivator.activate returns the failure case. This causes `show` to render, but now that oversub report policy returns true and the view errors since the variables weren't prepared.

Whew.

Adding `if: ->() { lock_intent_changed? }` fixes the issue (and I've confirmed that it still runs when you try to lock intent).

app/views/draws/_form.html.erb
```diff
@@ -1,6 +1,7 @@
 <%= simple_form_for @draw do |f| %>
   <%= f.input :name %>
   <%= f.input :intent_deadline %>
+  <%= f.input :lock_intent %>
```
orenyk: Do we want to make this a button near the intent report instead? We don't even have a way for admins to navigate to `/draws/:id/edit` at the moment.

spec/features/users/intent_spec.rb
```diff
@@ -2,17 +2,37 @@
...
-    visit "users/#{student.id}/intent"
-    select('off_campus', from: 'user_intent')
-    click_on 'Submit Intent'
+  context 'admin' do
+    let!(:draw) { FactoryGirl.create(:draw, status: 'pre_lottery') }
+    it 'can lock intent changes on a per draw basis' do
+      student = FactoryGirl.create(:student, draw: draw)
+      lock_intent(draw)
+      log_in student
+      expect(page).not_to have_content('Update Housing Intent')
```
orenyk: I was going to ask where this link lives but then I realized it's in the nav. Maybe add a comment for future us' sake?

app/models/draw.rb
esoterik: oh wow it really shows that i did this quickly while in class haha

esoterik:
hmm, right now the update intent shows only when a draw is `pre_lottery` -- did you mean that we should lock intent when we move to `lottery`?

I agree with your other comments and will update this later!

orenyk:
Gotcha - I mean, if we're already checking draw status in the intent policy then we can leave it, but ultimately we _could_ use the `lock_intent` boolean as the only determination for that policy and simply toggle it when we switch draw statuses. No need to update the service objects for now though, if you think it's still relevant you can open a separate issue. Thanks!

orenyk:
On second thought (as I work on #314), I think we _should_ update `DrawLotteryStarter` to include the locking of intent - this can then be used in the policy for bulk intent setting (although admins will conceivably still be able to make individual intent updates).

orenyk:
Should we also rename the attribute to `intent_locked` for clarity? `lock_intent` sounds more like a method.

esoterik:
this should also edit the flash message of setting all intents to on campus to include a link to lock intents

esoterik:
@orenyk ready to rereview

app/models/draw.rb
```diff
@@ -139,4 +142,11 @@ def group_count
   def remove_old_draw_ids
     User.where(old_draw_id: id).update_all(old_draw_id: nil)
   end
+
+  def cannot_lock_intent_if_undeclared
+    undeclared = students.where(intent: 'undeclared')
+    return if undeclared.empty?
```
orenyk: Can't we use `all_intents_declared?` here?

app/policies/user_policy.rb
```diff
@@ -34,4 +34,10 @@ def resolve
       scope
     end
   end
+
+  private
+
+  def draw_intent_state
+    !user.draw.intent_locked && !user.draw.draft? if user.draw
```
orenyk: So this is meant to return false / nil if the user is not in a draw? I'd almost rather:

```ruby
def draw_intent_state
  return false unless user.draw
  !user.draw.intent_locked && !user.draw.draft?
end
```

I'm still a little concerned about letting students change intent DURING a lottery - at the very least we should validate that the student isn't already in a group / doesn't have any memberships before allowing them to switch from on_campus (or at the very least destroy all non-accepted memberships if they try to).

Any thoughts?

orenyk: Wait, looking below it looks like you did implement (lock intent when we start the lottery) - if that's the case, can't we replace `!user.draw.draft?` with `user.draw.pre_lottery?`? Admins will still be able to update intents.

app/helpers/draws_helper.rb
```diff
@@ -22,6 +22,14 @@ def diff_class(diff)
     'zero'
   end
 
+  def lock_intent_btn_str(draw)
```
orenyk: Document?

app/helpers/draws_helper.rb
```diff
@@ -22,6 +22,14 @@ def diff_class(diff)
...
+    if draw.intent_locked
+      'Allow students to update their intent'
```
orenyk: These are _really_ long button labels - can we make them more concise and add tooltips instead?

app/services/draw_lottery_starter.rb
```diff
@@ -35,7 +35,7 @@ def initialize(draw:)
   #   message to set in the flash and either `nil` or the modified object.
   def start
     return error unless valid?
-    return success if draw.update(status: 'lottery')
+    return success if draw.update(status: 'lottery', intent_locked: true)
```
orenyk: :+1:

app/helpers/draws_helper.rb
esoterik: good catch! I also need to document the new attribute in the `Draw` class

app/policies/user_policy.rb
esoterik: that's what I had before and this is the problem I was running into yesterday -- the idea is that if admins want, they can allow students to change their intent in any phase. 

I can change this, but I figured more flexibility would be better?

orenyk: Sure that's fine, you can keep it more flexible.

esoterik:
@orenyk back to you!

app/views/draws/show.html.erb
```diff
@@ -19,40 +19,32 @@
...
+          input_html: { value: !@draw.intent_locked } %>
+        <%= f.submit lock_intent_btn_label(@draw),
+                     **with_tooltip(text: lock_intent_btn_tooltip(@draw)) %>
+      <% end %>
+    <% end %>
+    <% if policy(@draw).bulk_on_campus? %>
+      <%= link_to 'Make all students on campus', bulk_on_campus_draw_path(@draw),
+                  method: :patch,
+                  **with_tooltip(text: 'Sets all students who have not declared'\
+                                 ' their housing intent to on campus',
```
orenyk: Nitpick (I know this came from my issue earlier) - `their housing intent to live on campus` - just while we're changing this line.

app/views/draws/show.html.erb
```diff
@@ -19,40 +19,32 @@
   <%= render 'ungrouped_report' %>
 </div>
 <% if policy(@draw).intent_summary? %>
-  <div class="intent-metrics">
-    <h3>Intent Metrics</h3>
-    <table>
-      <tbody>
-        <% @intent_metrics.each do |intent_value, count| %>
```
orenyk: If we're getting rid of this we should remove the controller code that calculates the metrics and the query object as well.

app/views/draws/show.html.erb
orenyk: Or maybe `their housing intent as living on campus`?

app/views/draws/show.html.erb
```diff
@@ -19,34 +19,32 @@
   <%= render 'ungrouped_report' %>
 </div>
 <% if policy(@draw).intent_summary? %>
```
orenyk: I don't even know if we need this policy or the `Intent Actions` headline, but it's your call.

Issue #224, opened 2017-02-21
Admins can disband oversubscribed groups following confirmation


orenyk:
Add an oversubscription handling page (linked to during the confirmation in #183) that allows admins to disband select groups in oversubscribed sizes.

1. This can be a page listing all groups broken down by size with a button to disband (JS confirmation, policy to show).
2. We can highlight oversubscribed sizes (background color).
3. We should also have a button to lock a specific suite size (policy-based).
4. List # of groups and # of suites for each size at the start of each section

orenyk:
Issue 1: view with group summary and group/suite counts (and highlight)
Issue 2: group disband button
Issue 3: lock size button

orenyk:
Still to do - add lock size button / route and separate oversubscription handling page (per suite size)

orenyk:
Ok, new view set up, but the current disband button takes you to the root path which is really bad. I'm trying to set up an AJAX destroy request that simply reloads the page, but it's not currently working. Calling it a night, will pick this up tomorrow.

orenyk:
Wow this was painful, opening the PR now.

PR #344, opened 2017-03-08
[209] Add transition to suite selection for draws

### LINKED ISSUE #209: Admins can activate suite selection phase for draws
Once all groups have lottery numbers, admins can proceed to suite selection which will notify the first group(s) that they are eligible to select their suite.

orenyk:
Resolves #209
- Add `Draw#lottery_complete?`
- Fix bug with `Draw#ungrouped_students?` (it now only looks at undeclared and on_campus students
- Add start_selection route and `DrawSelectionStarter` service object

### DIFF
--- a/app/models/draw.rb
+++ b/app/models/draw.rb
@@ -6,8 +6,8 @@
 # @attr students [Array<User>] The students in the draw.
 # @attr suites [Array<Suite>] The suites in the draw.
 # @attr status [String] The status / phase of the draw (draft, pre_lottery,
-#   lottery, post_lottery). Note the use of underscores in the status strings;
-#   this prevents some unpleasantness with the helper methods.
+#   lottery, suite_selection). Note the use of underscores in the status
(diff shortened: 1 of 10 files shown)


app/models/draw.rb
```diff
@@ -69,7 +69,8 @@ def groups?
   #
   # @return [Boolean] whether or not there are any ungrouped students
   def ungrouped_students?
-    students.includes(:group).select { |s| s.group.nil? }.count.positive?
+    students.where(intent: %w(on_campus undeclared)).includes(:group)
+            .select { |s| s.group.nil? }.count.positive?
```
esoterik: should this still use the query object?

esoterik: also, undeclared students should be handled separately

orenyk: We're going to move this to #224, merging!

orenyk:
Rebased, merging after green build (ignoring CodeClimate).

PR #345, opened 2017-03-09
[343] Add notification for the next groups to select a suite

### LINKED ISSUE #343: Notify group leaders when it's their turn to pick a suite
Following #209, the service object that updates the draw status to suite_selection will hit a second service object that e-mails the next available group leaders by lottery number. We will then also use that service object when group leaders select suites (blocked on #210)

orenyk:
Resolves #343
- Add NextGroupsQuery
- Add Draw#next_groups
- Add notification to DrawSelectionStarter

Blocked on #344


orenyk:
Rebased, no longer blocked and ready for review!

esoterik:
actually those are my only comments, back to you!

PR #347, opened 2017-03-09
[346] Fix _inline_checkboxes_override stylesheet

### LINKED ISSUE #346: Move / fix inline checkbox override stylesheet
The final version is in the root directory and an older version is in `app/assets/stylesheets`... fixing now.

orenyk:
Resolves #346
- Moves the correct version from the root directory into app/assets/

Not sure how this happened but should be fine to merge.


orenyk:
Oh damn it was me, I approved and merged #125 😞

PR #340, opened 2017-03-07
[224] Add oversubscription handling to draws

orenyk:
Resolves #224
- Redirect lottery start requests to oversubscription page if draw is oversubscribed, but can be bypassed even while oversubscribed
- Add Draw#oversubscribed? and Draw#size_locked?
- Add GroupSizesQuery
- Add DrawSizesQuery (yay Arel)
- Allow group disbanding requests to take a path to redirect to
- Add draw size locking route (also takes a redirect path via params)
- Add DrawsHelper#toggle_size_lock_btn

PR #364, opened 2017-03-09
[314] Add bulk assignment of draw students to on-campus intent

### LINKED ISSUE #314: Admins should be able to set the intent of all students in a draw to On Campus

orenyk:
Resolves #314
- Add BulkOnCampusUpdater
- Add Draw#before_lottery?

### DIFF
--- a/app/controllers/draws_controller.rb
+++ b/app/controllers/draws_controller.rb
@@ -53,6 +53,14 @@ class DrawsController < ApplicationController # rubocop:disable ClassLength
     render action: 'intent_report'
   end
 
+  def bulk_on_campus
+    result = BulkOnCampusUpdater.update(draw: @draw)
+    # note that BulkOnCampusUpdater.update always returns a success hash with
+    # :object set to @draw, so we don't need to handle a fallback case via
+    # handle_action
+    handle_action(**result)
+  end
+
   def suite_summary
     @all_sizes = SuiteSizesQuery.new(Suite.available).call
     @suites_by_size = SuitesBySizeQuery.new(@draw.suites.available).call
--- a/app/policies/draw_policy.rb
+++ b/app/policies/draw_policy.rb
@@ -46,6 +46,10 @@ class DrawPolicy < ApplicationPolicy
     !record.draft?
   end
 
+  def bulk_on_campus?
+    edit? && record.before_lottery? && !record.all_intents_declared?
+  end
+
   def oversub_report?
     !record.draft? && !record.suites.empty?
   end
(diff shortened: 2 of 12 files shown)


app/controllers/draws_controller.rb
```diff
@@ -3,7 +3,8 @@
 # Controller for Draws
 class DrawsController < ApplicationController # rubocop:disable ClassLength
   prepend_before_action :set_draw, except: %i(index new create)
-  before_action :calculate_metrics, only: %i(show activate start_lottery)
+  before_action :calculate_metrics, only: %i(show activate start_lottery
+                                             bulk_on_campus)
```
esoterik: we don't need to calculate this for `bulk_on_campus`

app/controllers/draws_controller.rb
```diff
@@ -52,6 +53,11 @@ def filter_intent_report
     render action: 'intent_report'
   end
 
+  def bulk_on_campus
+    result = BulkOnCampusUpdater.update(draw: @draw)
+    handle_action(action: 'show', **result)
```
esoterik: remove the action and note that this always returns a "success"

orenyk:
Done! @esoterik ready for re-review, I rebased to deal with the first round of merge conflicts but this is also going to conflict with #340, so that's going to be fun 😞.

app/views/draws/show.html.erb
```diff
@@ -37,6 +37,9 @@
     <% if policy(@draw).intent_report? %>
       <p><%= link_to 'View intent report', intent_report_draw_path(@draw) %></p>
     <% end %>
+    <% if policy(@draw).bulk_on_campus? %>
+      <p><%= link_to 'Make all students on-campus', bulk_on_campus_draw_path(@draw), method: :patch, class: 'button' %></p>
+    <% end %>
```
esoterik: this should live outside of the intent report div so it can be seen in draft draws.

app/policies/draw_policy.rb
```diff
@@ -46,6 +46,10 @@ def intent_summary?
     !record.draft?
   end
 
+  def bulk_on_campus?
+    edit? && record.before_lottery?
+  end
```
esoterik: should this also check that there are students with undeclared intents?

app/views/draws/show.html.erb
esoterik: also, should we add a descriptive [tooltip](http://foundation.zurb.com/sites/docs/v/5.5.3/components/tooltips.html)?

I think this one should say "Sets all students who haven't declared their housing intent to on campus"

Issue #348, opened 2017-03-09
Rework suite assignment for draws


esoterik:
One page / form, split up by suite size (convert `suites_edit.html.erb` to a partial), with check/uncheck all (woo JS, see #178 for details).

orenyk:
This should also make sure not to include medical suites in the form.

orenyk:
merging #178 into here

PR #370, opened 2017-03-12
[326] Add medical suite flag

### LINKED ISSUE #326: Add medical suite tag for suites
Just a boolean attribute. Open questions:

1. What happens with merging / splitting?

orenyk:
Resolves #326
- Add boolean :medical attribute to Suites
- Add hidden form on Suite show to toggle the medical attribute
- Add SuitesHelper with methods to generate the form and describe a given suite

I'm actually liking the pattern for `simple_form_for resource` with hidden fields so we don't need separate routes and service objects for simple changes; obviously it requires validations to live on the model so it doesn't always work, but we may be able to simplify a bit when we refactor.

### DIFF
--- a/app/helpers/suites_helper.rb
+++ b/app/helpers/suites_helper.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+#
+# Helper module for Suites
+module SuitesHelper
+  # Sets the value for the hidden field on the medical suite "form"
+  #
+  # @param suite [Suite] the suite in question
+  # @return [Integer] a value inverse of the current medical status
+  def medical_val(suite)
+    suite.medical ? 0 : 1
+  end
+
+  # Returns the button string for the medical suite "form"
+  #
+  # @param suite [Suite] the suite in question
+  # @return [String] the button string
+  def medical_btn_str(suite)
+    "Make #{medical_str(!suite.medical)}"
+  end
+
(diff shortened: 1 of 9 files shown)


app/helpers/suites_helper.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+#
+# Helper module for Suites
+module SuitesHelper
+  # Sets the value for the hidden field on the medical suite "form"
+  #
+  # @param suite [Suite] the suite in question
+  # @return [Integer] a value inverse of the current medical status
+  def medical_val(suite)
```
esoterik: can we name this so that it's more obvious that this is for the hidden field?

orenyk:
I will:

1. Prevent medical suites from being displayed in Buildings#show tables (next in the `each` loop)
2. Check medical suite status in the policy for show
3. Prevent medical suites from being assigned to draws (likely in the form collection, to be handled in #348.

app/helpers/suites_helper.rb
```diff
@@ -0,0 +1,29 @@
...
+    suite.medical ? 0 : 1
```
esoterik: this can just be `!suite.medical` so it doesn't need to be a helper

orenyk:
@esoterik back to you!

PR #371, opened 2017-03-12
[329] Handle IDR timeouts

### LINKED ISSUE #329: Gracefully handle IDR timeout issues

orenyk:
Resolves #329
- ~~This currently rescues the exception in UserBuilder, but this might be dangerous as multiple failures during an Enrollment wouldn't be caught (so we might end up with very very long requests that don't exit). See [here](https://github.com/heroku/rack-timeout#errors) for more details (particularly the bullet point for RequestTimeoutException)~~
- Rescues the timeout exception in the controllers to avoid multiple timeouts during enrollment.


orenyk:
@esoterik back to you! Added feature specs as well.

PR #379, opened 2017-03-14
[146] Allow students to leave groups

### LINKED ISSUE #146: Students should be able to leave non-frozen groups

orenyk:
Resolves #146
- Add :leave route and policy for groups in draws

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -76,6 +76,11 @@ class GroupsController < ApplicationController
     handle_action(path: draw_group_path(@draw, @group), **result)
   end
 
+  def leave
+    result = MembershipDestroyer.destroy(membership: current_user.membership)
+    handle_action(path: draw_group_path(@draw, @group), **result)
+  end
+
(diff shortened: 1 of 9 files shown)


app/controllers/groups_controller.rb
```diff
@@ -76,6 +76,16 @@ def reject_pending
     handle_action(path: draw_group_path(@draw, @group), **result)
   end
 
+  def leave
+    membership = current_user.membership
+    result = if membership
```
esoterik: why is this necessary if we check for this in the policy?

orenyk: Hmm, that's a good question. I guess we don't - if the current user is in `group.members` it _must_ have an accepted membership. I'll update this, thanks!

orenyk:
@esoterik ended up tweaking the MembershipDestroyer a bit, want to take a quick look?

esoterik:
ready to rebase and merge!

orenyk:
Rebased, will merge after green build

PR #383, opened 2017-03-14
[382] Add Confirmation before proceeding to lottery

### LINKED ISSUE #382: Add Proceeding to Lottery confirmation page
Right now, it's too easy to accidentally proceed to lottery phase, and there's no way to go back, so we need an intermediate page that shows:
* The draw summary table (gives us group metrics, inventory metrics, oversubscription metrics)
* The students without groups summary
* Some warning text about what happens if you proceed without resolving oversubscription if there's oversubscription
* A link to the handle oversubscription page if there's oversubscription

esoterik:
Resolves #382

### DIFF
--- a/app/controllers/draws_controller.rb
+++ b/app/controllers/draws_controller.rb
@@ -4,7 +4,8 @@
 class DrawsController < ApplicationController # rubocop:disable ClassLength
   prepend_before_action :set_draw, except: %i(index new create)
   before_action :calculate_metrics, only: %i(show activate start_lottery
-                                             start_selection)
+                                             start_selection
+                                             lottery_confirmation)
 
   def show; end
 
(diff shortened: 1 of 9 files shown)


app/controllers/draws_controller.rb
```diff
@@ -99,10 +100,12 @@ def students_update
     handle_action(**result)
   end
 
+  def lottery_confirmation; end
+
   def start_lottery
     @lottery_starter = DrawLotteryStarter.new(draw: @draw)
     result = @lottery_starter.start
-    handle_action(action: 'show', **result)
+    handle_action(path: lottery_confirmation_draw_path(@draw), **result)
```
orenyk: Why wouldn't use `action: 'lottery_confirmation'`? You're already calculating the metrics for this action as well.

app/views/draws/lottery_confirmation.html.erb
```diff
@@ -0,0 +1,23 @@
...
+    <%= render 'oversub_report', draw: @draw %>
+  </div>
+  <hr />
+<% end %>
+<div class="ungrouped-report">
+  <%= render 'ungrouped_report' %>
+</div>
+<hr />
+<% if @draw.oversubscribed? %>
+  <h3> Warning! </h3>
```
orenyk: Unnecessary spaces.

app/views/draws/lottery_confirmation.html.erb
```diff
@@ -0,0 +1,23 @@
...
+  There are still oversubscribed suite sizes. If you choose to proceed to
```
orenyk: This should be surrounded by a `<p>` tag.

Issue #278, opened 2017-03-03
Draw summary tables should include group sizes without suites


orenyk:
In the DrawsController, instead of populating the oversub_report and the group_report based simply on the available suite sizes, we should also include the sizes of all groups in the draw (even if there are no longer available suites left).

orenyk:
This should be trivial after #209 is merged in since that includes the DrawSizesQuery, we'd just have to update the controller logic a bit.

orenyk:
Nvm this is blocked on #224 since that's where DrawSizesQuery ended up.

esoterik:
resolved by #384

PR #384, opened 2017-03-14
[88] Disband groups with unavailable suite sizes on starting lottery

### LINKED ISSUE #88: Handle group sizes when a suite size becomes unavailable
Group sizes are restricted to suite sizes available in the draw at the time of group creation/editing, but those sizes may become unavailable after the group sets its size.

We need to figure out how to handle this situation. I think that an email notification to the affected groups that tells them they have to change their size along with a banner notification for the group leader while signed in would work well.

esoterik:
Resolves #88

### DIFF
--- a/app/controllers/draws_controller.rb
+++ b/app/controllers/draws_controller.rb
@@ -163,33 +163,39 @@ class DrawsController < ApplicationController # rubocop:disable ClassLength
   end
 
   def calculate_metrics
+    calculate_sizes
     calculate_suite_metrics
     calculate_group_metrics
     calculate_oversub_metrics
     calculate_ungrouped_students_metrics
   end
 
+  def calculate_sizes
+    @suite_sizes ||= @draw.suite_sizes
+    @group_sizes ||= @draw.group_sizes
+    @sizes ||= (@suite_sizes + @group_sizes).uniq.sort
+  end
+
   def calculate_suite_metrics
-    @suite_sizes = SuiteSizesQuery.new(@draw.suites.available).call
     @suite_counts = @draw.suites.available.group(:size).count
+    @suite_counts.default = 0
   end
 
   def calculate_group_metrics
-    @draw_sizes = DrawSizesQuery.call(draw: @draw)
-    empty_groups_hash = @draw_sizes.map { |s| [s, []] }.to_h
     @groups = @draw.groups.includes(:leader)
-    @groups_by_size =
-      empty_groups_hash.merge(@groups.sort_by { |g| Group.statuses[g.status] }
-                                     .group_by(&:size))
+    @groups_by_size = @groups.sort_by { |g| Group.statuses[g.status] }
+                             .group_by(&:size)
+    @groups_by_size.default = []
   end
 
-  def calculate_oversub_metrics # rubocop:disable AbcSize
+  def calculate_oversub_metrics
     return unless policy(@draw).oversub_report?
-    zeroed_group_hash = @suite_sizes.map { |s| [s, 0] }.to_h
-    @group_counts = zeroed_group_hash.merge(@groups.group(:size).count)
-    @locked_counts = zeroed_group_hash.merge(@groups.where(status: 'locked')
-                                                    .group(:size).count)
-    @diff = @suite_sizes.map do |size|
+    calculate_sizes
+    @group_counts = @groups.group(:size).count
+    @group_counts.default = 0
+    @locked_counts = @groups.where(status: 'locked').group(:size).count
+    @locked_counts.default = 0
+    @diff = @sizes.map do |size|
       [size, @suite_counts[size] - @group_counts[size]]
     end.to_h
   end
--- a/app/helpers/draws_helper.rb
+++ b/app/helpers/draws_helper.rb
@@ -14,7 +14,7 @@ module DrawsHelper
   end
 
   def diff_class(diff)
-    raise ArgumentError unless diff.is_a? Integer
+    return '' unless diff.is_a? Integer
     return 'positive' if diff.positive?
     return 'negative' if diff.negative?
     'zero'
--- a/app/models/draw.rb
+++ b/app/models/draw.rb
@@ -34,6 +34,13 @@ class Draw < ApplicationRecord
     SuiteSizesQuery.new(suites.available).call
   end
 
+  # Finds the sizes that groups exist for within the draw
+  #
+  # @return [Array<Integer>] the available group sizes
+  def group_sizes
+    @group_sizes ||= GroupSizesQuery.new(groups).call
+  end
+
   # Finds all suite sizes for which new groups can be created by removing
   # locked_sizes from available sizes
   #
@@ -134,7 +141,7 @@ class Draw < ApplicationRecord
   #
   # @return [Boolean] whether or not the draw is oversubscribed
   def oversubscribed?
-    @oversubscribed ||= GroupSizesQuery.new(groups).call.any? do |size|
+    @oversubscribed ||= group_sizes.any? do |size|
       groups.where(size: size).count > suites.where(size: size).count
     end
   end
--- a/app/views/draws/_oversub_report.html.erb
+++ b/app/views/draws/_oversub_report.html.erb
@@ -11,7 +11,7 @@
     </tr>
   </thead>
   <tbody>
-    <% @suite_sizes.each do |size| %>
+    <% @sizes.each do |size| %>
       <tr class="<%= diff_class(@diff[size])%>">
         <td data-role="suite-size"><%= size_str(size).capitalize.pluralize %></td>
         <td data-role="suite-count"><%= @suite_counts[size] %></td>
--- a/spec/models/draw_spec.rb
+++ b/spec/models/draw_spec.rb
@@ -44,6 +44,17 @@ RSpec.describe Draw, type: :model do
     end
   end
 
+  describe '#group_sizes' do
+    let(:draw) { FactoryGirl.build_stubbed(:draw) }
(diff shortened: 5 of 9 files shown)


app/models/draw.rb
```diff
@@ -34,6 +34,13 @@ def suite_sizes
     SuiteSizesQuery.new(suites.available).call
   end
 
+  # Finds the sizes that groups exist for within the draw
+  #
+  # @return [Array<Integer>] the available group sizes
+  def group_sizes
```
orenyk: We should memoize this and use it in `oversubscribed?`.

orenyk:
Bigger issue - if you want to actually set up this scenario then draw#show doesn't render. Steps to replicate:

1. Add group of a given size to a draw
2. Remove all suites of that size from draw
3. Return to draw#show

It's probably because of #278, so maybe we just merge that into here?

esoterik:
@orenyk  ready for review! now resolves #278 as well as #88

app/controllers/draws_controller.rb
```diff
@@ -183,15 +184,17 @@ def calculate_group_metrics
...
       empty_groups_hash.merge(@groups.sort_by { |g| Group.statuses[g.status] }
                                      .group_by(&:size))
+    @groups_by_size.default = 0
+    @group_sizes = @groups_by_size.keys
   end
 
   def calculate_oversub_metrics # rubocop:disable AbcSize
     return unless policy(@draw).oversub_report?
-    zeroed_group_hash = @suite_sizes.map { |s| [s, 0] }.to_h
+    zeroed_group_hash = @group_sizes.map { |s| [s, 0] }.to_h
```
orenyk: I feel like we want to use `@group_sizes + @suite_sizes` here, otherwise we won't show suite sizes with no groups. I think that was originally my point in #278, sorry if it wasn't clear.

spec/models/draw_spec.rb
```diff
@@ -44,6 +44,17 @@
...
   end
 
+  describe '#group_sizes' do
+    let(:draw) { FactoryGirl.build_stubbed(:draw) }
+    it 'returns an array of group sizes in the draw' do
+      instance_spy('GroupSizesQuery', call: [1, 2]).tap do |q|
+        allow(GroupSizesQuery).to receive(:new).with(draw.suites.available)
+          .and_return(q)
+      end
+      expect(draw.group_sizes).to match_array([1, 2])
```
orenyk: We can just assert that `GroupSizesQuery` gets called, we don't need to test against return values.

spec/models/draw_spec.rb
```diff
@@ -44,6 +44,17 @@
     end
   end
 
+  describe '#group_sizes' do
+    let(:draw) { FactoryGirl.build_stubbed(:draw) }
+    it 'returns an array of group sizes in the draw' do
+      instance_spy('GroupSizesQuery', call: [1, 2]).tap do |q|
+        allow(GroupSizesQuery).to receive(:new).with(draw.suites.available)
```
orenyk: Snake case, just to be safe.

app/controllers/draws_controller.rb
orenyk: Also, just to be clear, this makes `calculate_oversub_metrics` dependent on `calculate_group_metrics`. It's fine, but we might want to comment that somewhere?

app/helpers/draws_helper.rb
```diff
@@ -14,7 +14,7 @@ def intent_deadline_str(draw)
   end
 
   def diff_class(diff)
-    raise ArgumentError unless diff.is_a? Integer
+    return 'negative' unless diff.is_a? Integer
```
orenyk: ... I don't like this. I _am_ ok with returning `''` if we don't pass an integer.

app/controllers/draws_controller.rb
```diff
@@ -174,6 +174,7 @@ def calculate_metrics
   def calculate_suite_metrics
     @suite_sizes = SuiteSizesQuery.new(@draw.suites.available).call
     @suite_counts = @draw.suites.available.group(:size).count
+    @suite_counts.default = 0
```
orenyk: Oo awesome - does this mean we can rid of `zeroed_group_hash` and `empty_groups_hash`? Didn't realize this was a thing, very cool.

app/helpers/draws_helper.rb
esoterik: oh, yeah this isn't even necessary any more -- I'll fix it

app/controllers/draws_controller.rb
esoterik: the oversub report still shows suites with no groups as-is

esoterik: it also shows sizes for which there are no suites but there are groups

orenyk: Oh ok, I'm not sure how I didn't see that but :+1:

orenyk: Ohhh, now I see. Ok, I think that `@group_sizes` is a confusing name then - it implies that it's only the group sizes. Can we make it clearer?

app/controllers/draws_controller.rb
```diff
@@ -165,36 +165,39 @@ def set_draw
...
-    @group_counts = zeroed_group_hash.merge(@groups.group(:size).count)
-    @locked_counts = zeroed_group_hash.merge(@groups.where(status: 'locked')
-                                                    .group(:size).count)
-    @diff = @group_sizes.map do |size|
+    calculate_sizes
+    @group_counts = @groups.group(:size).count
+    @group_counts.default = 0
+    @locked_counts = @groups.where(status: 'locked').group(:size).count
+    @locked_counts.default = 0
+    @diff = @sizes.map do |size|
```
orenyk: Minor,  but do we want to replace this with `each_with_object` to avoid `to_h`? Now that we've found it I think it reads much better.

spec/features/draws/draw_oversubscription_handling_spec.rb
```diff
@@ -23,6 +23,8 @@
     expect(page).to have_content('Oversubscription handling')
   end
 
+  # TODO: figure out why this spec fails to send sizes to the
```
orenyk: Is this still broken?

spec/models/draw_spec.rb
```diff
@@ -47,11 +47,12 @@
   describe '#group_sizes' do
     let(:draw) { FactoryGirl.build_stubbed(:draw) }
     it 'returns an array of group sizes in the draw' do
-      instance_spy('GroupSizesQuery', call: [1, 2]).tap do |q|
+      g = instance_spy('group_sizes_query', call: [1, 2]).tap do |q|
```
orenyk: We don't need to specify a return for `call` here.

app/views/draws/_group_report.html.erb
```diff
@@ -1,7 +1,8 @@
 <% if groups.all? { |_s, groups| groups.empty? } %>
   <p>There are no groups yet.</p>
 <% else %>
-  <% sizes.each do |size| %>
+  <% # THIS IS THE HACKY FIX %>
```
orenyk: ?? Was this fixing the issue in the spec before?

app/views/draws/_oversub_report.html.erb
```diff
@@ -11,7 +11,7 @@
     </tr>
   </thead>
   <tbody>
-    <% @suite_sizes.each do |size| %>
+    <% @sizes.each do |size| %>
```
orenyk: Are we not currently sending in local variables to this partial? Fine to leave if not.

spec/models/draw_spec.rb
```diff
@@ -44,6 +44,18 @@
     end
   end
 
+  describe '#group_sizes' do
+    let(:draw) { FactoryGirl.build_stubbed(:draw) }
+    it 'returns an array of group sizes in the draw' do
+      g = instance_spy('group_sizes_query', call: [1, 2]).tap do |q|
```
orenyk: I don't think we need to specify the return of `call` here anymore.

orenyk:
Also, rubocop 😛

app/views/draws/_group_report.html.erb
orenyk: Found it - you need to update `app/views/draws/oversubscription.html.erb` as well to pass `@sizes` instead of `@draw_sizes`.

spec/features/draws/draw_oversubscription_handling_spec.rb
orenyk: See [my other comment](https://github.com/YaleSTC/vesta/pull/384#discussion_r106812467).

spec/models/draw_spec.rb
esoterik: I'm not sure that we can just test that the method was called on the query object here -- we be ensuring that the return value of `group_sizes` is correct, so we need to set a return value for `call`

app/controllers/draws_controller.rb
esoterik: I think it's ok for now, plus we will be refactoring this later

spec/features/draws/draw_oversubscription_handling_spec.rb
orenyk: Can we remove this comment?

spec/models/draw_spec.rb
```diff
@@ -161,18 +172,6 @@
     end
   end
 
-  describe '#all_intents_declared?' do
```
orenyk: Why did you remove this test?

esoterik: it was a duplicate

orenyk:
PR failure is intermittent and fixed in #410, merging.

Issue #99, opened 2017-01-12
Add College(?) model for configuration


orenyk:
We should add a model, like `ResCollege` to store configuration params for each instance / tenant. This will include things like the college name, the admin e-mail, the dean's name, etc. Beginning with #51, we've added places where this would be necessary so here's a list of places to update in this issue:

* [ ] Default `from` e-mail address in `ApplicationMailer` (maybe?)
* [ ] Default `from` e-mail address in `StudentMailer`
* [ ] The `@res_college` instance variable in the `StudentMailer.draw_invitation` method / views. Update this variable name to match the model name.
* TBD

orenyk:
This will likely need to be expanded in #303

orenyk:
Tried some fancy stuff with Ruby delegation to turn it into a Singleton class for now but that didn't work, will get back to this tomorrow.

PR #386, opened 2017-03-15
[348] [178] Rework suite assignment for draws

### LINKED ISSUE #178: Add "Check-All" button to Update Suites on Draws
Blocked on #175 

The form that updates suites in draws should include a check all button

### LINKED ISSUE #348: Rework suite assignment for draws

orenyk:
Resolves #348, resolves #178
- Consolidate all suite sizes onto one form
- Add method_missing to DrawSuiteUpdate to set dynamic attributes for each suite size
- Update DrawsController to consolidate the params from the form
- Add check/uncheck all links for each size on the form (woo JS magic)

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -21,3 +21,18 @@ $(function(){ $(document).foundation(); });
 $(document).on('change', '.autosubmitme', function() {
   $(this).parents('form').submit();
 });
+
+$(document).on('click', '.suites-fields-section .check-all', function() {
+  set_checkbox_of_parent(this, '.suites-fields-section', true);
+  return false; // prevent scrolling behavior
+});
+
+$(document).on('click', '.suites-fields-section .uncheck-all', function() {
+  set_checkbox_of_parent(this, '.suites-fields-section', false);
+  return false; // prevent scrolling behavior
+});
+
+function set_checkbox_of_parent(element, parent_class, checked) {
+  var parent_div = $(element).parents(parent_class)[0];
+  $(parent_div).find('input[type="checkbox"]').prop('checked', checked);
+}
--- a/app/controllers/draws_controller.rb
+++ b/app/controllers/draws_controller.rb
@@ -143,9 +143,7 @@ class DrawsController < ApplicationController # rubocop:disable ClassLength
   end
 
   def suites_update_params
-    params.require(:draw_suites_update).permit(:size, suite_ids: [],
-                                                      drawn_suite_ids: [],
-                                                      undrawn_suite_ids: [])
+    params.require(:draw_suites_update).permit(suite_edit_param_hash)
   end
 
   def students_update_params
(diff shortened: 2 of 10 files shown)


app/assets/javascripts/application.js
```diff
@@ -21,3 +21,18 @@ $(function(){ $(document).foundation(); });
 $(document).on('change', '.autosubmitme', function() {
   $(this).parents('form').submit();
 });
+
+$(document).on('click', '.suites-fields .check-all', function() {
+  set_checkbox_of_parent(this, '.suites-fields', true);
+  return false; // prevent scrolling behavior
```
orenyk: Apparently returning false in the other function doesn't carry through to the callback (as I would expect in Ruby).

app/models/draw_suites_update.rb
```diff
@@ -4,7 +4,7 @@
 class DrawSuitesUpdate
   include ActiveModel::Model
 
-  attr_reader :size, :suite_ids, :drawn_suite_ids, :undrawn_suite_ids
+  attr_reader :suite_ids, :drawn_ids, :drawless_ids
```
orenyk: I'm actually not sure we even need these readers at this point.

app/models/draw_suites_update.rb
```diff
@@ -105,4 +104,27 @@ def error(error)
...
+    @all_suite_sizes ||= SuiteSizesQuery.new(Suite.available).call
+  end
+
+  def params_for(size)
+    param_headings = %w(suite_ids drawn_ids drawless_ids)
+    param_headings.map { |p| "#{p}_#{size}".to_sym }
+  end
+
+  def respond_to_missing?(method_name, include_all = false)
+    valid_method_names.include?(method_name) || super
```
esoterik: what does `super` hit here?

app/controllers/draws_controller.rb
```diff
@@ -243,4 +249,33 @@ def process_student_assignment
...
+
+  def suite_edit_param_headings
+    %w(suite_ids drawn_ids drawless_ids)
+  end
+
+  # This method is necessary to consolidate the suite update params from all the
+  # different suite sizes into three distinct params - this could be moved to
+  # the DrawSuitesUpdate service object in the future since we're already
+  # setting up getters for all of the possible param instance variables, but
+  # this could just as easily live in the controller.
```
esoterik: I think this should go in the service object

app/controllers/draws_controller.rb
```diff
@@ -199,14 +198,21 @@ def calculate_ungrouped_students_metrics
     @ungrouped_students.delete('off_campus')
   end
 
-  def prepare_suites_edit_data # rubocop:disable AbcSize
+  def prepare_suites_edit_data # rubocop:disable AbcSize, MethodLength
```
esoterik: this makes me feel like we need a whole separate `DrawSuitesController` but leaving it alone for now

orenyk: We definitely do.

app/models/draw_suites_update.rb
orenyk: It hits the default behavior in `Object` I believe, which is `false` unless it shows up in the Ruby method inheritance chain.

app/controllers/draws_controller.rb
orenyk: Ok, I'll move it and request another review.

orenyk:
@esoterik back to you!

orenyk:
Oh sorry forgot about that change, updating now.

orenyk:
Also rebasing on #99 since that will be next to merge, I think.

orenyk:
@esoterik back to you, just verify that you like how it looks / works and we should be good to merge.

PR #388, opened 2017-03-15
[313] Add ability to unlock group memberships

### LINKED ISSUE #313: Admins should be able to edit a locked single

orenyk:
Resolves #313
- Add :unlock routes / actions to groups
- Add GroupUnlocker service object
- Tweak membership validations to permit unlocking of memberships
- Tweak Group#update_status! to revert status from locked
- Update GroupPolicy#lock? to permit the locking of groups in statuses other than full (e.g. finalizing)
- Update GroupUpdater#update to run Group#update_status! after other updates are made
- Delegate :lock? and :unlock? methods in DrawlessGroupPolicy to GroupPolicy to keep things DRY

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -95,6 +95,13 @@ class Group < ApplicationRecord # rubocop:disable ClassLength
     (members - locked_members).empty? && members_count == size
   end
 
+  # Check if there are any locked memberships
+  #
+  # @return [Boolean] true if there are any unlocked members
+  def unlockable?
+    !locked_members.empty? && suite.nil?
+  end
+
   private
 
   # override default attribute getter to include transfers
@@ -163,7 +170,7 @@ class Group < ApplicationRecord # rubocop:disable ClassLength
     if members_count < size
       self.status = 'open'
     elsif members_count == size
-      self.status = 'full' if open?
+      self.status = 'full' unless finalizing? || lockable?
       self.status = 'locked' if lockable?
     end
   end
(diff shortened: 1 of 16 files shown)


app/models/group.rb
```diff
@@ -95,6 +95,13 @@ def lockable?
     (members - locked_members).empty? && members_count == size
   end
 
+  # Check if there are any locked memberships
+  #
+  # @return [Boolean] true if there are any unlocked members
+  def unlockable?
+    !locked_members.empty?
```
esoterik: should this also check that the group doesn't have a suite?

app/services/updaters/group_updater.rb
```diff
@@ -59,6 +58,13 @@ def delete_leader_id_if_empty
   end
 
   # Note that this occurs within the transaction
+  def add_and_remove_users
+    remove_users if pending_users[:remove]
+    group.reload
```
esoterik: reloading to update the status?

app/services/updaters/group_updater.rb
```diff
@@ -59,6 +58,13 @@ def delete_leader_id_if_empty
   end
 
   # Note that this occurs within the transaction
+  def add_and_remove_users
```
esoterik: `update_members` is a better name for this method I think

app/models/group.rb
orenyk: Sure, that makes sense.

app/services/updaters/group_updater.rb
orenyk: Yea, just to be safe.

app/services/updaters/group_updater.rb
orenyk: :+1:

Issue #276, opened 2017-03-03
Admins/reps can assign suites to groups in draws with lottery numbers


orenyk:
Fix the policy when we write the policy in #210

orenyk:
@esoterik should this include the ability for admins to remove groups from draws?

orenyk:
This will be a separate interaction:

1. Admins / reps can view a page / form with _all_ groups currently eligible for suite selection and dropdowns of the appropriate suites.
2. They select suites for all of the eligible groups and submit.
3. We then redirect to the same view but we now present the next group(s) for selection.

PR #390, opened 2017-03-16
[210] Add ability for leaders to select suites

### LINKED ISSUE #210: Group Leader can select suites
When it's a group's turn to pick a suite, the leader should be able to select an appropriately sized suite. 

After the group selects a suite, an email should be sent to the next group.

There might be some stuff we can copy from the drawless group suite assignment.

esoterik:
Resolves #210
  - Add GroupSuiteSelector

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -5,6 +5,8 @@ class GroupsController < ApplicationController
   layout 'application_with_sidebar'
   prepend_before_action :set_group, except: %i(new create)
   prepend_before_action :set_draw
+  before_action :authorize_draw!, except: %i(select_suite assign_suite show
+                                             assign_lottery)
   before_action :set_form_data, only: %i(new edit)
 
   def show
@@ -113,6 +115,16 @@ class GroupsController < ApplicationController
     handle_action(action: 'show', **result)
   end
 
+  def select_suite
+    @suites = @draw.suites.available.where(size: @group.size)
+  end
+
+  def assign_suite
+    suite_id = group_params['suite']
+    result = GroupSuiteSelector.select(group: @group, suite_id: suite_id)
+    handle_action(action: 'select_suite', **result)
+  end
+
   private
 
   def authorize!
@@ -121,14 +133,17 @@ class GroupsController < ApplicationController
     else
       authorize Group
     end
+  end
+
+  def authorize_draw!
     authorize @draw, :group_actions?
   end
 
   def group_params
     p = params.require(:group).permit(:size, :leader_id, :transfers,
-                                      :lottery_number, member_ids: [],
-                                                       remove_ids: [],
-                                                       invitations: [])
+                                      :lottery_number, :suite,
+                                      member_ids: [], remove_ids: [],
+                                      invitations: [])
     return p if @group
     p.reject! { |k, _v| k == 'transfers' }
   end
--- a/app/models/draw.rb
+++ b/app/models/draw.rb
@@ -159,7 +159,24 @@ class Draw < ApplicationRecord
   #
   # @return [Array<Group>] the next available groups using NextGroupsQuery
   def next_groups
-    NextGroupsQuery.call(draw: self)
+    @next_groups ||= NextGroupsQuery.call(draw: self)
+  end
+
+  # Returns true if the given group is in the next groups to pick suites
+  #
+  # @return [Boolean]
+  def next_group?(group)
+    next_groups.include? group
+  end
+
+  # Email the leaders of the next groups to select suites.
+  #
+  # @param [Mailer] The mailer to use
+  def notify_next_groups(mailer = StudentMailer)
+    college = College.first
+    next_groups.each do |group|
+      mailer.selection_invite(group.leader, college).deliver_later
+    end
   end
 
   private
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -88,6 +88,13 @@ class User < ApplicationRecord
     intent.tr('_', ' ')
   end
 
+  # Returns true if the user is the leader of the given group
+  #
+  # @return [Boolean]
+  def leader_of?(group)
+    self == group.leader
+  end
+
   # Back up a user's current draw into old_draw_id and removes them from current
   # draw, also setting intent to undeclared. Does nothing if draw_id is nil.
   #
--- a/app/policies/group_policy.rb
+++ b/app/policies/group_policy.rb
@@ -99,6 +99,14 @@ class GroupPolicy < ApplicationPolicy
     user.admin? && rooms_assigned?(record)
   end
 
+  def select_suite?
+    user.admin? || (user.leader_of?(record) && record.draw.next_group?(record))
+  end
+
+  def assign_suite?
+    select_suite?
+  end
+
   class Scope < Scope # rubocop:disable Style/Documentation
     def resolve
       scope
--- a/app/services/group_suite_selector.rb
+++ b/app/services/group_suite_selector.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+#
+# Service objects for group leaders to select a suite for their group.
+class GroupSuiteSelector < SuiteSelector
+  # Initialize a new GroupSuiteSelector
+  #
+  # @param group [Group] the group to assign the suite to
+  # @param suite_id [Integer] the suite id to assign to the group
+  # @param mailer [Mailer] the mailer to use to email next groups
+  def initialize(group:, suite_id:, mailer: StudentMailer)
+    @mailer = mailer
+    super(group: group, suite_id: suite_id)
+  end
+
+  # Select / assign a suite to a group. Checks to make sure that the group does
+  # not curretly have a suite assigned, that the suite_id corresponds to an
+  # existing suite, and that the suite is not currently assigned to a different
+  # group. Notifies next groups to select suites.
+  #
+  # @return [Hash{symbol=>Group,Hash}] a results hash with a message to set in
+  #   the flash, nil or the group as the :object, and an action to render.
+  def select
+    if assign_suite_to_group
+      notify_next_groups
+      success
+    else
+      error
+    end
+  end
+
+  private
+
+  attr_reader :mailer
+
+  def draw
+    group.draw
+  end
+
+  def notify_next_groups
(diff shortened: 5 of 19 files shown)


app/controllers/groups_controller.rb
```diff
@@ -103,6 +105,16 @@ def assign_lottery
     @color_class = @group.save ? 'success' : 'failure'
   end
 
+  def select_suite
+    @suites = @draw.suites.where(size: @group.size)
```
orenyk: `@draw.suites.available`

app/policies/group_policy.rb
```diff
@@ -83,6 +83,15 @@ def change_leader?
     edit?
   end
 
+  def select_suite?
+    user.admin? || (user.leader_of?(record) &&
+                    record.draw.next_groups.any? { |g| g == record })
```
orenyk: Why don't we move this to a model method on `draw`, e.g. `Draw#next_group?`?

app/services/group_suite_selector.rb
```diff
@@ -0,0 +1,53 @@
...
+  # @param suite_id [Integer] the suite id to assign to the group
+  # @param mailer [Mailer] the mailer to use to email next groups
+  def initialize(group:, suite_id:, mailer: StudentMailer)
+    @group = group
+    @suite_id = suite_id
+    @errors = []
+    @mailer = mailer
+  end
+
+  # Select / assign a suite to a group. Checks to make sure that the grou does
```
orenyk: typo, `grou`

spec/features/groups/suite_selection_spec.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.feature 'Suite Selection' do
+  let(:leader) do
+    FactoryGirl.create(:draw_in_selection).next_groups.first.leader.tap do |l|
+      l.update(password: 'password')
```
orenyk: ?

spec/services/group_suite_selector_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+        suite = mock_suite(id: 123, present: false)
+        described_class.select(group: group, suite_id: suite.id.to_s)
+        expect(group.draw).not_to have_received(:notify_next_groups)
+      end
+      it 'sets an error message in the flash' do
+        draw = instance_spy('Draw', next_groups: [])
+        group = instance_spy('group', suite: nil, draw: draw)
+        suite = mock_suite(id: 123, present: false)
+        result = described_class.select(group: group, suite_id: suite.id.to_s)
+        expect(result[:msg].keys).to include(:error)
```
orenyk: I thought you wanted to check exclusivity here, e.g. `match([:error])`?

spec/services/group_suite_selector_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    end
+    context' success' do
+      # rubocop:disable RSpec/ExampleLength
+      it 'sends emails to the next groups' do
+        next_group = instance_spy('Group', lottery_number: 12)
+        draw = instance_spy('Draw', next_groups: [next_group])
+        group = instance_spy('group', suite: nil, draw: draw)
+        suite = mock_suite(id: 123)
+        described_class.select(group: group, suite_id: suite.id.to_s)
+        expect(group.draw).to have_received(:notify_next_groups)
```
orenyk: Do we want to check that it passes `next_group.leader`?

spec/services/group_suite_selector_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+        group = instance_spy('group', suite: nil, draw: draw)
+        suite = mock_suite(id: 123, present: false)
+        result = described_class.select(group: group, suite_id: suite.id.to_s)
+        expect(result[:msg].keys).to include(:error)
+      end
+    end
+    context' success' do
+      # rubocop:disable RSpec/ExampleLength
+      it 'sends emails to the next groups' do
+        next_group = instance_spy('Group', lottery_number: 12)
```
orenyk: Why does this need a lottery number?

spec/services/group_suite_selector_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+        draw = instance_spy('Draw', next_groups: [next_group])
+        group = instance_spy('group', suite: nil, draw: draw)
```
orenyk: both `group` and `suite` could be set in `let` blocks to make the tests more concise.

spec/services/group_suite_selector_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+        suite = mock_suite(id: 123)
+        result = described_class.select(group: group, suite_id: suite.id.to_s)
+        expect(result[:object]).to eq([group.draw, group])
+      end
+      it 'sets a success message in the flash' do
+        draw = instance_spy('Draw', next_groups: [])
+        group = instance_spy('group', suite: nil, draw: draw)
+        suite = mock_suite(id: 123)
+        result = described_class.select(group: group, suite_id: suite.id.to_s)
+        expect(result[:msg].keys).to include(:success)
```
orenyk: same here, `match` vs `include`.

app/policies/group_policy.rb
```diff
@@ -83,6 +83,15 @@ def change_leader?
     edit?
   end
 
+  def select_suite?
```
orenyk: So this policy will change in #276? Or should we restrict selection to groups that are the next groups in their respective draws?

spec/models/draw_spec.rb
```diff
@@ -255,10 +255,23 @@
...
     end
   end
+
+  describe '#notify_next_groups' do
+    # rubocop:disable RSpec/ExampleLength
+    it 'sends a selection invite to the leaders of the next groups' do
+      draw = FactoryGirl.build_stubbed(:draw)
+      group = instance_spy('Group', leader: instance_spy('User'))
+      allow(draw).to receive(:next_groups).and_return([group])
+      mailer = instance_spy('StudentMailer')
```
orenyk: ummm... changing this to `instance_spy('student_mailer')` passes this test... not sure what's going on. I'll leave it to you to update and rebase.

spec/features/groups/suite_selection_spec.rb
esoterik: no idea but we can't log in with the user otherwise.

spec/services/group_suite_selector_spec.rb
esoterik: oh yeah you're right

app/policies/group_policy.rb
esoterik: I think it would change -- reps can assign as long as the group is in the next groups

spec/services/group_suite_selector_spec.rb
esoterik: realized that I used `include` because I copied this from `SuiteSelector` spec -- changing it there as well

esoterik: also we want `eq`, not `match`

spec/services/group_suite_selector_spec.rb
esoterik: it only sends emails when the lottery number is different than the lottery number of the current group

esoterik:
@orenyk ready for rereview! will squash and rebase after this one

esoterik:
no, i reverted them because that's the file that caused the segfault when i changed it; i'll see what i can do to avoid the segfault and fix them as much as possible -- but i'm not super concerned because it's essentially just cleaning up tests

orenyk:
Oh ok, I wasn't sure. It's fine to just skip it all if you want then if those changes are causing problems - there will still be the merge conflicts to address.

orenyk:
Failing test? Also, might as well rebase and solve merge conflicts before a final review.

app/models/draw.rb
```diff
@@ -152,7 +152,24 @@ def size_locked?(size)
...
+  # @return [Boolean]
+  def next_group?(group)
+    next_groups.include? group
+  end
+
+  # Email the leaders of the next groups to select suites.
+  #
+  # @param [Mailer] The mailer to use
+  def notify_next_groups(mailer = StudentMailer)
+    college = College.first
```
orenyk: This can just be `@current_college` I think.

app/models/user.rb
```diff
@@ -88,6 +88,13 @@ def pretty_intent
     intent.tr('_', ' ')
   end
 
+  # Returns true if the user is the leader of the given group
+  #
+  # @return [Boolean]
+  def leader_of?(group)
```
orenyk: Ooo I feel like we could heavily refactor `GroupPolicy` and `DrawlessGroupPolicy`, if not `DrawPolicy. I'll open an issue for post-pilot.

app/services/group_suite_selector.rb
```diff
@@ -0,0 +1,53 @@
...
+#
+# Service objects for group leaders to select a suite for their group.
+class GroupSuiteSelector < SuiteSelector
+  # Initialize a new GroupSuiteSelector
+  #
+  # @param group [Group] the group to assign the suite to
+  # @param suite_id [Integer] the suite id to assign to the group
+  # @param mailer [Mailer] the mailer to use to email next groups
+  def initialize(group:, suite_id:, mailer: StudentMailer)
+    @group = group
```
orenyk: Can we use `super` here instead for the first three instance variables?

app/views/groups/select_suite.html.erb
```diff
@@ -0,0 +1,5 @@
+<h1>Select a Suite</h1>
+<%= simple_form_for [@draw, @group], url: assign_suite_draw_group_path(@draw, @group) do |f| %>
+  <%= f.input :suite, collection: @suites, as: :radio_buttons, label_method: :number %>
+  <%= f.submit "Submit Selection" %>
```
orenyk: Do we want to add a tooltip here? Maybe something about how this is final and commits them to living on campus?

spec/services/group_suite_selector_spec.rb
orenyk: Can we be consistent about the capitalization of class names as passed to `instance_spy`? I think we should stick with camel case since that fixed that weird test before, but either way we should be consistent within a given test.

PR #392, opened 2017-03-16
[99] Add Colleges

### LINKED ISSUE #99: Add College(?) model for configuration

orenyk:
Resolves #99
- Add College model with basic e-mail attributes
- Add college routes / actions
- Add factory
- Replace structs in mailers with DB records
- Add CollegeGenerator and college generation to the seed script
- Add settings links to the nav

This will likely be the table that is used for multi-tenancy when we implement that. It will also likely conflict with #390 since it changes where the e-mail is triggered from, but it should be an easy-enough conflict to resolve (in either direction).

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,6 +4,7 @@ class ApplicationController < ActionController::Base
   include Pundit
   protect_from_forgery with: :exception
   before_action :authenticate_user!, unless: :devise_controller?
+  before_action :check_college, unless: :devise_controller?
   before_action :authorize!, except: :home, unless: :devise_controller?
   after_action :verify_authorized, except: :home, unless: :devise_controller?
 
@@ -50,5 +51,15 @@ class ApplicationController < ActionController::Base
   # @abstract
   def authorize!
     raise NoMethodError
+  end
+
+  def check_college
+    return if current_college.persisted? || !policy(College).edit?
+    flash[:warning] = 'You must update your college settings in order for '\
+      'Vesta to function correctly'
+  end
+
+  def current_college
+    @current_college ||= College.first || College.new
   end
 end
--- a/app/models/college.rb
+++ b/app/models/college.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+#
+# Model for site configuration / individual colleges. Will be used for
+# multi-tenancy in the future.
+#
+# @attr name [String] the college name
+# @attr admin_email [String] the admin e-mail for the college
+# @attr dean [String] the name of the college dean / head
+# @attr site_url [String] the full url (including http/https) for the Vesta site
+#   for the college
+class College < ApplicationRecord
+  validates :name, presence: true, uniqueness: { case_sensitive: false }
+  validates :admin_email, presence: true
+  validates :dean, presence: :true
+  validates :site_url, presence: :true
+end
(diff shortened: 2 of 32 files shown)


app/controllers/application_controller.rb
```diff
@@ -51,4 +52,15 @@ def complete_request(action: nil, path: nil)
   def authorize!
     raise NoMethodError
   end
+
+  def check_college
```
esoterik: I know this is what we did in reservations but I really don't like it -- especially because it means that we need to create this model in all of our feature specs.

I would rather have some sort of persistent nag / a form on the home page to update these settings until they are set?

app/models/college.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+#
+# Model for site configuration / individual colleges. Will be used for
+# multi-tenancy in the future.
+#
+# @attr name [String] the college name
+# @attr admin_email [String] the admin e-mail for the college
+# @attr dean [String] the name of the college dean / head
+# @attr site_url [String] the full url (including http/https) for the Vesta site
+#   for the college
```
esoterik: also need housing rep emails and link to housing rules

we should also note in the "student experience" issue that we need to add these things to this model:
* box link to floorplans
* some sort of static text that is displayed to students

(noting this here & will comment on the issue when I'm done)

app/views/colleges/index.html.erb
```diff
@@ -0,0 +1,8 @@
+<h1> Inventory </h1>
```
esoterik: remove the unnecessary views from buildings

orenyk: done

app/models/college.rb
orenyk: I figured we'd add this stuff in the other issue, this was more to just replace what we already had.

I also don't know that we need to store housing rep e-mails separately - we'll know who they are by their user accounts and roles. In fact, we can just create `Draw#reps` and use that to display the e-mail address in the draw views.

app/controllers/application_controller.rb
orenyk: Ok, below set `@current_college` to `College.first || College.new` and add a flash message if it isn't `persisted`.

orenyk:
@esoterik back to you!

app/controllers/application_controller.rb
```diff
@@ -51,4 +52,14 @@ def complete_request(action: nil, path: nil)
...
+    return if current_college.persisted? || !policy(College).edit?
```
esoterik: why are we checking the edit policy here?

orenyk: Because why show the flash to students / reps who won't be able to do anything about it?

orenyk:
@esoterik I'm merging pending green build, responded to your comment above.

Issue #321, opened 2017-03-05
Add unauthenticated homepage with general information (help page, maybe?)


orenyk:
We should have the root path take you to a page accessible to unauthenticated users with general information about Vesta. This should fix the whole "logging in with CAS still shows an error flash message" issue as well.

esoterik:
probably should just say something like "This is the housing draw app for X. If you think you should have access to this site, contact Y." ?

Issue #303, opened 2017-03-03
[META] Dashboards for Students/Housing Reps/Admins


AlanLiu96:
Feedback from User Testing:

Housing Reps mentioned that they'd like to have these features on the dashboard. I separated them by the groups the suggestions applied to.

All:
- [ ] Deadlines
   - Intent, Group Formation, Draw

Users:
- [ ] Instructions
- [ ] Floorplans
- [ ] Available rooms for current group size

Admin/Housing Rep:
- [ ] Group-less Students, Undeclared Students

orenyk:
So a lot of this information is draw-by-draw - are housing reps assigned to specific class years / draws?

AlanLiu96:
By assignment, I'm not sure if you're referring to being a member of the draw or organizing/managing that draw. The answer though is the same in either case.

@spaceraccoon Correct me if I'm mistaken, but I believe that there are one or two members of the housing committee per draw. (Ex. 2 Sophomores for sophomore draw) 

They would both run the draws they're assigned to and be participants of the draw.

orenyk:
Ok that's more what I was getting at - whether or not housing committee members need to see / help manage more than one draw. @esoterik this might play into our discussion of housing rep permissions in Pundit - we may want to check against their membership in a given draw.

orenyk:
We need to take into account room selection data when present for students and reps. Example here:

```erb
<% if current_user.room.present? %>
  <h1>Your housing information</h1>
  <h3>Suite: <%= @group.suite.number %></h3>
  <h3>Room: <%= @room.number_with_type %></h3>
  <h3>Group: </h3>
  <ul>
    <% @group.members.each do |m| %>
      <li><%= m.full_name %> (<%= m.room.number %>)</li>
    <% end %>
  </ul>
<% else %>
  <h1>Vesta</h1>
<% end %>
```

PR #396, opened 2017-03-17
[211] Room selection

### LINKED ISSUE #211: Students can select their rooms
This will technically happen offline, but group leaders can enter the room selections for their entire group. A second issue will be opened for a more comprehensive feature.

orenyk:
:boom:

Resolves #211 
- Add RoomAssignment singular resource / service object
- Lots of spec / policy updates

### DIFF
--- a/app/controllers/room_assignments_controller.rb
+++ b/app/controllers/room_assignments_controller.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+#
+# Controller for room assignments
+class RoomAssignmentsController < ApplicationController
+  prepend_before_action :set_room_assignment
+  prepend_before_action :set_group
+  before_action :set_rooms, only: %i(new edit)
+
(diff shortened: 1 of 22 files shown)


esoterik:
@orenyk I fixed a failing spec, changed the dropdown slightly, and added a display of housing information on the student homepage after rooms are assigned. can you look this over before we merge?

app/views/application/home.html.erb
```diff
@@ -1 +1,13 @@
-<h1>Vesta</h1>
+<% if current_user.room.present? %>
```
orenyk: So... this is basically the student dashboard, right, or at least the start of one (i.e. #303)? I feel like this route will end up being our unauthenticated home page (e.g. #321) and/or will be switched to using HighVoltage, so I'm not sure why it goes in this issue, tbh.

PR #397, opened 2017-03-17
[323] Add export of group leader emails

### LINKED ISSUE #323: Allow admins / reps to export a list of group leader e-mails
1. All groups
2. All unlocked groups
3. All locked groups of a given suite size

Which one(s) do we want to add?

### LINKED ISSUE #410: User destruction broken for non-leaders
changing the membership destruction test to destroy a non leader:
```ruby
  it 'destroys a dependent membership on destruction' do
    user = FactoryGirl.create(:student, intent: 'on_campus')
    g = FactoryGirl.create(:drawless_group, size: 2)
    m_id = Membership.create(group: g, user: user).id
    user.destroy
    expect { Membership.find(m_id) }.to \
      raise_error(ActiveRecord::RecordNotFound)
  end
```
results in this error
```
  1) User destroys a dependent membership on destruction
     Failure/Error: user.destroy

     ActiveRecord::StatementInvalid:
       PG::UndefinedColumn: ERROR:  column groups.user_id does not exist
       LINE 1: SELECT  "groups".* FROM "groups" WHERE "groups"."user_id" = ...
                                                      ^
       : SELECT  "groups".* FROM "groups" WHERE "groups"."user_id" = $1 LIMIT $2
     # ./spec/models/user_spec.rb:47:in `block (2 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # PG::UndefinedColumn:
     #   ERROR:  column groups.user_id does not exist
     #   LINE 1: SELECT  "groups".* FROM "groups" WHERE "groups"."user_id" = ...
     #                                                  ^
     #   ./spec/models/user_spec.rb:47:in `block (2 levels) in <top (required)>'
```

orenyk:
Resolves #323
- Add email_export singular resource, with policy and helper

Also resolves #410

### DIFF
--- a/app/models/email_export.rb
+++ b/app/models/email_export.rb
@@ -1,0 +1,68 @@
+# frozen_string_literal: true
+#
+# Service / form object for email exports
+#
+# @attr draw_id [Integer] the draw_id to scope to, if any; set to zero for
+#   drawless groups only
+# @attr size [Integer,Nil] the size to scope to, if any
+# @attr locked [Boolean] whether or not to scope to locked groups only
+class EmailExport
+  include ActiveModel::Model
+
+  attr_accessor :draw_id, :size, :locked
+  attr_reader :leaders
+
+  validates :draw_id, numericality: { only_integer: true, allow_nil: true,
+                                      greater_than_or_equal_to: 0 }
+  validates :size, numericality: { only_integer: true, allow_nil: true,
+                                   greater_than: 0 }
+  validates :locked, inclusion: { in: [true, false] }
+
+  # Initialize a new EmailExport, nillifies empty strings for draw_id and
+  # converts locked to actual booleans
+  def initialize(*args)
+    super(*args)
+    process_locked
+    process_draw_id
+  end
+
+  # Generate the e-mail export, collects the relevant groups and their leaders
+  #
+  # @return [EmailExport] the email export object, will have a leaders attribute
+  #   populated with the query results if it succeeded
+  def generate
+    execute_query if valid?
+    self
+  end
+
+  # Determine whether or not the query was scoped to a specific draw or to
+  # drawless groups
+  #
+  # @return [Boolean] whether or not the export was scoped to draw_id
+  def draw_scoped?
+    draw_scope || false
+  end
+
+  private
+
+  attr_reader :draw_scope
+
+  def process_locked
+    self.locked = locked.to_i == 1 ? true : false
+  end
+
+  def process_draw_id
+    return unless draw_id.present?
+    @draw_scope = true
+    self.draw_id = nil if draw_id.to_i.zero?
+  end
+
+  def execute_query # rubocop:disable AbcSize
+    query = User.includes(:led_group).where.not(groups: { id: nil })
+                .order(:last_name, :first_name)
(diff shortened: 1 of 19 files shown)


app/models/email_export.rb
```diff
@@ -0,0 +1,69 @@
...
+  def process_locked
+    self.locked = locked.to_i == 1 ? true : false
+  end
+
+  def process_draw_id
+    return unless draw_id.present?
+    @draw_scope = true
+    # this kinda makes the numericality validations on :draw_id useless, not
+    # sure if it's worth keeping them or not.
+    self.draw_id = nil if draw_id.to_i.zero?
```
esoterik: in what situations would this actually happen? I think the validations would prevent this, right?

app/models/email_export.rb
```diff
@@ -0,0 +1,69 @@
...
+  end
+
+  def execute_query # rubocop:disable AbcSize
+    query = User.includes(:led_group).where.not(groups: { id: nil })
+    query = query.where(groups: { draw_id: draw_id }) if draw_scope
+    query = query.where(groups: { size: size }) if size.present?
+    query = query.where(groups: { status: Group.statuses['locked'] }) if locked
```
esoterik: we should refactor these queries into scopes so we can build the chain we want and then send it to `User` all at once instead of potentially running 4 queries in this method

orenyk: It's not executing the query until we call `.to_a` at the end, I'm pretty sure. I think ActiveRecord is smart about that until you treat a relation like an Enumerable.

app/models/email_export.rb
orenyk: This is used to scope for drawless groups - we need to set it to `nil` from `0` but differentiate that from the case where no `draw_id` is passed (in which case we don't scope by draw). However, if you pass _any_ value other than a positive integer in, `to_i` will turn it into zero regardless. In other words, if you param hack to send `draw_id = 'a'`, that will still scope to drawless groups.

That said, I now disagree with my comment - the numericality validation on `draw_id` is _not_ useless since it prevents the queries from borking if someone called `execute_query` after assigning some non-integer value to `draw_id` (resulting in an invalid query). I'm going to remove the comment (but copy it below for posterity):

```ruby
    # this kinda makes the numericality validations on :draw_id useless, not
    # sure if it's worth keeping them or not.
```

orenyk: This is used to scope for drawless groups - we need to set it to `nil` from `0` but differentiate that from the case where no `draw_id` is passed (in which case we don't scope by draw). However, if you pass _any_ value other than a positive integer in, `to_i` will turn it into zero regardless. In other words, if you param hack to send `draw_id = 'a'`, that will still scope to drawless groups.

That said, I now disagree with my comment - the numericality validation on `draw_id` is _not_ useless since it prevents the queries from borking if someone called `execute_query` after assigning some non-integer value to `draw_id` (resulting in an invalid query). I'm going to remove the comment (but copy it below for posterity):

```ruby
    # this kinda makes the numericality validations on :draw_id useless, not
    # sure if it's worth keeping them or not.
```

orenyk:
@esoterik responded to both of your comments (one of them got hidden in the rebase) and added rep navigation that will likely need to be reworked. Let me know what you think!

PR #402, opened 2017-03-19
[387] Prevent admins from creating groups in draft draws

### LINKED ISSUE #387: Admins shouldn't be able to create groups in draft draws
Simple policy update

orenyk:
Resolves #387
- Update DrawPolicy#group_actions?

### DIFF
--- a/app/views/draws/show.html.erb
+++ b/app/views/draws/show.html.erb
@@ -69,7 +69,7 @@
   <% if policy(@draw).lottery? %>
     <%= link_to 'Assign lottery numbers', lottery_draw_path(@draw), class: 'button secondary' %>
   <% end %>
-  <% if policy(Group).new? %>
+  <% if policy(Group).new? && policy(@draw).group_actions? %>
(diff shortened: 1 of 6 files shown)


app/views/draws/show.html.erb
```diff
@@ -69,7 +69,7 @@
   <% if policy(@draw).lottery? %>
     <%= link_to 'Assign lottery numbers', lottery_draw_path(@draw), class: 'button secondary' %>
   <% end %>
-  <% if policy(Group).new? %>
+  <% if policy(Group).new? && policy(@draw).group_actions? %>
```
esoterik: I don't like this -- I think it should all live in the `new?` policy.

orenyk: There's no way to pass the `@draw` variable to the GroupPolicy for `new?`, since there's no actual group involved (it authorizes the class). I looked into this and it is strongly discouraged anyway, see [here](https://github.com/elabs/pundit#additional-context). I know it's not great but I'm not sure there's an alternative. Let me know what you think!

orenyk:
@esoterik responded, back to you!

PR #407, opened 2017-03-19
[405] Ensure that room assignments are cleared appropriately

### LINKED ISSUE #405: Clear user room_ids when necessary
1. When group suite assignment changes
2. When group is disbanded

orenyk:
Resolves #405
- Add before_save callback to Suite conditional on group_id changing
- Add before_destroy callback to Group

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -126,7 +126,7 @@ class User < ApplicationRecord
   end
 
   def room_in_suite
-    return if room.suite_id == group.suite.id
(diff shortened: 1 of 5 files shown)


app/models/user.rb
```diff
@@ -126,7 +126,7 @@ def cas_auth?
   end
 
   def room_in_suite
-    return if room.suite_id == group.suite.id
+    return if !room || room.suite_id == group.suite.id
```
esoterik: good catch

PR #409, opened 2017-03-20
[148] Ensure membership errors display correctly

### LINKED ISSUE #148: Membership Errors should be more specific
Right now, when group creation fails because the memberships are invalid, the only error shown is "Memberships invalid" which isn't helpful. We should figure out how to get the more specific membership errors into the flash.

orenyk:
Resolves #148
- Check membership validity before validating a group
- Change create / update actions to rendering responses instead of
  redirecting for groups
- Fix bug in GroupsController#create that prevented group errors from
  being displayed after invalid submission
- Tweak group creator to properly handle invalid params
- Fix all the ensuing bugs :-\

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -38,12 +38,10 @@ class Group < ApplicationRecord # rubocop:disable ClassLength
   validate :validate_status, if: ->(g) { g.size.present? }
 
   before_validation :add_leader_to_members, if: ->(g) { g.leader.present? }
+  after_save :update_status!, if: ->(g) { g.transfers_changed? }
   before_destroy :remove_member_rooms
-
   after_destroy :restore_member_draws, if: ->(g) { g.draw.nil? }
   after_destroy :notify_members_of_disband
-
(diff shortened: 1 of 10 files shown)


orenyk:
That's not actually a regression - it's broken on master. This is actually kinda expected - the size validation won't fail until the full memberships are all persisted, which can never happen. Not sure how to resolve this... we could add a separate non-ActiveModel validation for size instead, if you'd like. Thoughts?

app/models/group.rb
```diff
@@ -124,6 +126,15 @@ def add_leader_to_members
     members << leader unless members.include? leader
   end
 
+  # rubocop:disable RaiseArgs
+  def check_full_membership_validity
```
esoterik: why isn't this just the first validation? we can have it stop the callback chain and not validate further, i think, and that way we can customize the error message

orenyk: Ohh rather than raising an exception? Maybe, let's chat about it offline and figure it out from there.

PR #427, opened 2017-03-21
[330] Update housing rep permissions

### LINKED ISSUE #330: Figure out / update housing rep permissions

orenyk:
Resolves #330
- Adjust permissions as per discussions with Dean's Office
- Clean up cases of checking :new? and :create? on an instance instead
  of the class
- Remove :index routes for Rooms and Suites


orenyk:
I need to add `SuitePolicy#medical?` and fix up the suite display - there were some issues there.

PR #436, opened 2017-03-23
[307] Clean up DrawlessGroups#show

### LINKED ISSUE #307: Font sizing on suite display on drawlessgroup#show

orenyk:
Resolves #307
- Use h3 instead of h2
- Remove unnecessary helper method from DrawlessGroupsHelper
- Fix incorrect query in DrawlessGroupsController#show

### DIFF
--- a/app/views/drawless_groups/show.html.erb
+++ b/app/views/drawless_groups/show.html.erb
@@ -1,20 +1,22 @@
 <%= render partial: 'groups/group_details' %>
 <% if policy(DrawlessGroup.new(@group)).select_suite? %>
-  <h2><%= suite_str(@group) %></h2>
-  <% if @group.suite %>
+  <h3><%= suite_str(@group) %></h3>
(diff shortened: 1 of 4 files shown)


app/views/drawless_groups/show.html.erb
```diff
@@ -1,20 +1,20 @@
 <%= render partial: 'groups/group_details' %>
 <% if policy(DrawlessGroup.new(@group)).select_suite? %>
-  <h2><%= suite_str(@group) %></h2>
+  <h3><%= suite_str(@group) %></h3>
   <% if @group.suite %>
```
esoterik: should this actually be nested in the policy?

orenyk: Yea, it ends up being the prompt to select a suite, which is restricted to admins. We can add suite display as an alternative - this view def needs some help.

orenyk:
@esoterik updated!

PR #437, opened 2017-03-23
[359] Improve draw oversubscription highlighting

### LINKED ISSUE #359: Update colors on draw show page
Draw Summary:
* Rows with more suites than groups should be green
* "Even" rows (those with the same number of suites as groups) should be blue

Group Report:
* Make the oversubscription highlighting less intense

orenyk:
Resolves #359
- Add DrawsHelper#oversub_str
- Tweak colors


esoterik:
this should also probably stop oversubscription warnings from showing during the suite selection phase

PR #438, opened 2017-03-23
[276] Add admin / rep suite assignment to draws

### LINKED ISSUE #276: Admins/reps can assign suites to groups in draws with lottery numbers

orenyk:
Resolves #276
- Add BulkSuiteSelector service
- Add :select_suites and :assign_suites routes to draw

Not a huge fan of radio buttons vs dropdowns, but we can discuss. I left the testing code in for both - we'll obviously want to clean that up before merging.

### DIFF
--- a/app/policies/draw_policy.rb
+++ b/app/policies/draw_policy.rb
@@ -86,6 +86,14 @@ class DrawPolicy < ApplicationPolicy
     edit? && record.lottery? && record.lottery_complete?
   end
 
+  def select_suites?
+    record.suite_selection? && (user.admin? || user.rep?)
+  end
+
+  def assign_suites?
+    select_suites?
+  end
+
   class Scope < Scope # rubocop:disable Style/Documentation
     def resolve
       scope
--- a/app/services/suite_selector.rb
+++ b/app/services/suite_selector.rb
@@ -2,6 +2,8 @@
 #
 # Service object to select suites for a given group.
 class SuiteSelector
+  attr_reader :errors
+
   # Allow for :select to be called on the parent class
   def self.select(**params)
     new(**params).select
@@ -34,7 +36,7 @@ class SuiteSelector
 
   private
 
-  attr_accessor :errors
+  attr_writer :errors
   attr_reader :group, :suite_id, :suite
 
   def valid?
--- a/app/controllers/draws_controller.rb
+++ b/app/controllers/draws_controller.rb
@@ -129,6 +129,26 @@ class DrawsController < ApplicationController # rubocop:disable ClassLength
     handle_action(action: 'show', **result)
   end
 
+  def select_suites
+    @groups = @draw.next_groups
+    if @groups.empty?
+      @draw.update!(status: 'results')
+      flash[:success] = 'All groups have suites!'
+      redirect_to draw_path(@draw)
+    end
+    @suite_selector = BulkSuiteSelectionForm.new(groups: @groups)
+    draw_suites = @draw.suites.available.where(size: @groups.map(&:size))
+    @suites_by_size = SuitesBySizeQuery.new(draw_suites).call
+  end
+
+  def assign_suites
+    @groups = @draw.next_groups
+    @suite_selector = BulkSuiteSelectionForm.new(groups: @groups)
+    @suite_selector.prepare(params: suite_selector_params)
+    result = @suite_selector.submit
+    handle_action(**result, path: select_suites_draw_path(@draw))
+  end
+
   private
 
   def authorize!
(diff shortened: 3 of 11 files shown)


app/services/bulk_suite_selector.rb
```diff
@@ -0,0 +1,136 @@
...
+    selector.errors
+  end
+
+  def all_groups_selected
+    return if all_groups_selected?
+    errors.add(:base, 'You must select a suite for all groups')
+  end
+
+  def all_groups_selected?
+    params.select { |_k, v| !v.empty? }.count == groups.count
```
esoterik: `params.values.reject(&:empty?).count == groups.count` is more concise

app/services/bulk_suite_selector.rb
```diff
@@ -0,0 +1,136 @@
...
+  private
+
+  attr_reader :groups, :params
+
+  # note that this occurs in a transaction
+  def process_all_suite_selections
+    results = groups.each_with_object({}) do |g, hash|
+      hash[g] = process_suite_selection(g)
+    end
+    return unless results.values.any? { |r| !r.empty? }
```
esoterik: `return unless results.values.none?(&:empty?)`

app/services/bulk_suite_selector.rb
```diff
@@ -0,0 +1,136 @@
...
+    return if params.values.uniq == params.values
+    errors.add(:base, 'You must select different suites for each group')
+  end
+
+  def suites_in_correct_draw
+    return unless all_groups_selected?
+    return unless groups.any? do |g|
+      next false unless g.draw
+      suite = Suite.find_by(id: send(param_for(g)))
+      next false if suite&.draws&.include? g.draw
```
esoterik: I don't think I like the use of `&.` here -- can we do `g.draw.suites.available.include? suite` instead?

app/services/bulk_suite_selector.rb
```diff
@@ -0,0 +1,136 @@
...
+      true
+    end
+    errors.add(:base, 'All groups in draws must be assigned to suites in '\
+      'the same draw')
+  end
+
+  def errors_from_results(results_hash)
+    results_hash.each_with_object([]) do |(g, r), array|
```
esoterik: is there a less-complicated way to store these errors? 

if the refactoring would take too much time, we should at least rename `r` to `result` to clarify things

esoterik: i guess we should also question if it's necessary to actually display all of the individual errors?

i'm sort of okay with a generic "oops something went wrong in assigning {GROUP NAME}" for errors that should not be reachable from the frontend, though that might not be the best

app/services/bulk_suite_selector.rb
```diff
@@ -0,0 +1,136 @@
+# frozen_string_literal: true
+#
+# Service object to handle the bulk assignment of suites to groups. Ensures that
+# suites are available, groups don't have suites assigned, and optionally that
+# suites belong to the appropriate draw.
+class BulkSuiteSelector
```
esoterik: should this be a form object? i think all/most of the logic for that is already here, unless i'm missing something

app/services/suite_selector.rb
```diff
@@ -34,7 +36,7 @@ def select
 
   private
 
-  attr_accessor :errors
+  attr_writer :errors
```
esoterik: why was this necessary? can it just stay `attr_accessor`?

app/policies/draw_policy.rb
```diff
@@ -86,6 +86,14 @@ def start_selection?
     edit? && record.lottery? && record.lottery_complete?
   end
 
+  def select_suites?
+    record.suite_selection? && (user.admin? || user.rep?)
```
esoterik: `!user.student?`?

app/services/bulk_suite_selector.rb
orenyk: Ooh nice

app/services/bulk_suite_selector.rb
orenyk: You're right, it basically is a form object. I'll rename / move it as appropriate.

app/services/bulk_suite_selector.rb
orenyk: Needs to be `reject!` but good suggestion.

app/services/bulk_suite_selector.rb
orenyk: Much better.

app/services/bulk_suite_selector.rb
orenyk: These errors are kinda reachable from the front-end, and I'd rather be more verbose now and deal with it down the road when we figure out the frontend properly. I renamed the block variables for clarity.

app/services/suite_selector.rb
orenyk: I believe that this overwrites the public reader (if I'm remembering correctly).

app/policies/draw_policy.rb
orenyk: I'd rather be explicit with permissions than not.

orenyk:
@esoterik updated!

orenyk:
Not sure what you mean, let's discuss tomorrow.

Issue #431, opened 2017-03-22
Turn housing resources into nested routes


orenyk:
E.g. `config/routes.rb` should look something like:

```ruby
resources :buildings do
  resources :suites do
    resources :rooms
  end
end
```

We also need to make sure we're appropriately handling association dependency - I'm adding `dependent: :destroy` for suites / buildings, but rooms are set to `dependent: :nullify` because of suite splitting / merging. This could result in orphaned rooms if a building was destroy, so we should fix that.

Finally, we can clean up the controllers / forms since the associations will be implicit by the routes / params.

orenyk:
We should also add proper cancel links to `rooms/new` and `suites/new`, e.g.

```ruby
# app/views/rooms/new.html.erb

...

<%= link_to 'Cancel', suite_path(@suite) %>
```
```ruby
# app/views/rooms_controller.rb
class RoomsController < ApplicationController
  before_action :set_suite

  ...

  private

  def set_suite
    @suite = Suite.find(params[:suite_id])
  end
end
```

PR #440, opened 2017-03-23
[341] Add back / cancel links where necessary

### LINKED ISSUE #341: Add cancel / back links to all views and relevant pages
Currently many of our forms and pages still don't have a clear navigation backwards (e.g. edit forms don't go back to the relevant resource). This can just be a simple link but we should add them.

orenyk:
Resolves #341 

There are going to be so many merge conflicts 😢. I left TODO's on the room and suite new pages since we currently can't redirect back to parent resource (e.g. suite and room, respectively), but we will be able to once we properly nest resources in #431. At the moment I added cancel links that go back to the buildings index, but I'm not sure that even makes sense. Let me know what you think!


esoterik:
going to block this until most of the other PRs are merged

PR #442, opened 2017-03-23
[321] Add unauthenticated landing page

### LINKED ISSUE #321: Add unauthenticated homepage with general information (help page, maybe?)

orenyk:
Resolves #321

### DIFF
--- a/app/views/pages/home.html.erb
+++ b/app/views/pages/home.html.erb
@@ -1,2 +1,11 @@
-<center><h1>Welcome to Vesta</h1></center>
+<div class="homepage-hero">
+  <h1>Welcome to Vesta</h1>
+  <p>
+  Vesta is a web app that manages the housing process for Yale's residential colleges.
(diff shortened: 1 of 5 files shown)


app/views/pages/home.html.erb
```diff
@@ -1,2 +1,11 @@
-<center><h1>Welcome to Vesta</h1></center>
+<div class="homepage-hero">
+  <h1>Welcome to Vesta</h1>
+  <p>
+  Vesta is a web app that is used to manage housing assignments and draws, nominally in a university setting.
```
esoterik: "Vesta is a web app that manages the housing process for Yale's residential colleges." is probably better for pilot

esoterik:
minor comment, but we can also leave it

PR #444, opened 2017-03-23
[421] Add secondary group report for later-stage draws

### LINKED ISSUE #421: Add lottery / selection group report
Should replace the current group report and include the lottery number and the suite selected (if any). I think we should still group by size but we can order by lottery number.

orenyk:
Resolves #421
- Fix bug with room assignments


esoterik:
this also needs to be rebased so the build will pass

orenyk:
@esoterik updated and rebased!

PR #462, opened 2017-03-27
[248] Add floor_plan_url and student_info_text to colleges

orenyk:
Resolves #248

PR #468, opened 2017-03-27
[303] dashboards

esoterik:
ugly but working

### DIFF
--- a/app/controllers/dashboards_controller.rb
+++ b/app/controllers/dashboards_controller.rb
@@ -1,0 +1,77 @@
+# frozen_string_literal: true
+# Controller for Dashboards
+class DashboardsController < ApplicationController
+  before_action :authorize!
+
+  def show
+    admin_metrics unless current_user.student?
+    student_variables unless current_user.admin?
+  end
+
+  private
+
+  def admin_metrics
+    @draws = Draw.all.includes(:groups).sort_by(&:name)
+    @draw_hash = Hash.new { |hash, key| hash[key] = {} }
+    @draws.each do |d|
+      @draw_hash[d][:partial] = d.status
+      if %w(pre_lottery suite_selection results).include? d.status
+        send("#{d.status}_metrics".to_sym, d, @draw_hash[d])
+      end
+    end
+  end
+
+  def pre_lottery_metrics(draw, hash) # rubocop:disable MethodLength, AbcSize
+    suite_counts = draw.suites.available.group(:size).count
+    suite_counts.default = 0
+    groups = draw.groups.sort_by { |g| Group.statuses[g.status] }
+                 .group_by(&:size)
+    groups.default = []
+    group_counts = groups.transform_values(&:count)
+    group_counts.default = 0
+    sizes = (suite_counts.keys + group_counts.keys).uniq.sort
+    diff = sizes.map { |s| [s, suite_counts[s] - group_counts[s]] }.to_h
+
+    locked_counts = groups.transform_values { |v| v.select(&:locked?).size }
+    locked_counts.default = 0
+
+    hash[:suite_counts] = suite_counts
+    hash[:group_counts] = group_counts
+    hash[:locked_counts] = locked_counts
+    hash[:groups_by_size] = groups
+    hash[:sizes] = sizes
+    hash[:diff] = diff
+
+    hash[:intent] = IntentMetricsQuery.call(draw)
+  end
+
+  def suite_selection_metrics(draw, hash)
+    with_suites = draw.groups.joins(:suite)
+
+    with_suite_count = with_suites.count
+    no_suite_count = draw.groups.count - with_suite_count
+    no_room_count = with_suite_count - with_suites.joins(leader: :room).count
+
+    hash[:next_groups] = draw.next_groups
+    hash[:no_suite_count] = no_suite_count
+    hash[:with_suite_count] = with_suite_count
+    hash[:no_room_count] = no_room_count
+  end
+
+  def results_metrics(draw, hash)
+    with_suites = draw.groups.joins(:suite)
+    no_room_count = with_suites.count - with_suites.joins(leader: :room).count
+    hash[:no_room_count] = no_room_count
+  end
+
+  def student_variables
+    @college = College.first || College.new
+    group = current_user.group
+    @suite = group.suite if group.present?
+    @room = current_user.room if @suite.present?
+  end
+
+  def authorize!
+    authorize Dashboard
+  end
+end
--- a/app/views/dashboards/_suite_selection.html.erb
+++ b/app/views/dashboards/_suite_selection.html.erb
@@ -1,0 +1,8 @@
+<% if hash[:next_groups] %>
+  <p>Groups currently picking suites: <%= hash[:next_groups].map(&:name).join(', ') %></p>
+  <p><%= hash[:no_suite_count] %> groups still have to select suites</p>
+  <p><%= hash[:with_suite_count] %> groups have selected suites</p>
+<% else %>
+  <p>All groups have selected suites.</p>
+<% end %>
+<p><%= hash[:no_room_count] %> groups with suites still have to select rooms</p>
--- a/app/views/dashboards/show.html.erb
+++ b/app/views/dashboards/show.html.erb
@@ -1,0 +1,47 @@
+<% unless current_user.admin? %>
+  <div class="card">
+    <div class="card-divider">
+      <h4>Welcome to Vesta!</h4>
+    </div>
+    <div class="card-section">
+      <p>Use the links above to navigate your housing draw.</p>
+      <% unless @college.student_info_text.blank? %>
+        <p><%= @college.student_info_text %></p>
+      <% end %>
+      <% unless @college.floor_plan_url.blank? %>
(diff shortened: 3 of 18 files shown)


esoterik:
blocked on #458, #462

app/controllers/dashboards_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+    @draw_hash = Hash.new { |hash, key| hash[key] = {} }
+    @draws.each do |d|
+      @draw_hash[d][:partial] = d.status
+      if %w(pre_lottery suite_selection results).include? d.status
+        send("#{d.status}_metrics".to_sym, d, @draw_hash[d])
+      end
+    end
+  end
+
+  # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
```
orenyk: If you want to put this inline you can get rid of the `Metrics/` prefix for both cops.

app/controllers/dashboards_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+  def pre_lottery_metrics(draw, hash)
+    suite_counts = draw.suites.available.group(:size).count
+    suite_counts.default = 0
+    groups = draw.groups.sort_by { |g| Group.statuses[g.status] }
+                 .group_by(&:size)
+    groups.default = []
+    group_counts = groups.map { |s, g| [s, g.count] }.to_h
```
orenyk: This could be replaced with `each_with_object` if you want.

orenyk: Actually, why not `transform_values`?

app/controllers/dashboards_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+    group_counts.default = 0
+    sizes = (suite_counts.keys + group_counts.keys).uniq.sort
+    diff = sizes.map { |s| [s, suite_counts[s] - group_counts[s]] }.to_h
```
orenyk: `each_with_object`?

app/controllers/dashboards_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    locked_counts = draw.groups.where(status: 'locked').group(:size).count
```
orenyk: Could this just be a transformed version of `groups`? I feel like we could use `select` instead of querying again.

app/views/dashboards/_suite_selection.html.erb
```diff
@@ -0,0 +1,8 @@
+<% if hash[:next_groups] %>
+  <p>Groups currently picking suites: <%= hash[:next_groups].map(&:name).join(', ') %></p>
+  <p><%= hash[:no_suite_count] %> groups still have to select suites</p>
+  <p><%= hash[:with_suite_count] %> groups have selected suites</p>
+<% else %>
+  <p>All groups have selected suites.</p>
```
orenyk: In principle we'd never hit this but good to have a fallback.

app/views/dashboards/show.html.erb
```diff
@@ -0,0 +1,37 @@
...
+          <h4><%= link_to draw.name, draw_path(draw) %> (<%= draw.status %>)</h4>
+        </div>
+        <div class="card-section">
+          <%= render hash[:partial], draw: draw, hash: hash %>
+        </div>
+      </div>
+    <% end %>
+  <% end %>
+<% end %>
+<% unless current_user.admin? %>
```
orenyk: Should we combine these into `if... elsif...`?

app/views/draws/_group_report.html.erb
```diff
@@ -1,21 +1,21 @@
 <% if groups.all? { |_s, groups| groups.empty? } %>
   <p>There are no groups yet.</p>
 <% else %>
-  <div><%= render 'email_exports/hidden_form', draw: @draw, size: nil, btn_str: "Export all group leader emails" %></div>
+  <div><%= render 'email_exports/hidden_form', draw: draw, size: nil, btn_str: "Export all group leader emails" %></div>
```
orenyk: Aren't we nuking these in #458? And below.

lib/tasks/style_checker.rake
```diff
@@ -38,7 +38,7 @@ def passed?
 end
 
 def rubocop(files)
-  "rubocop -D --force-exclusion #{files}"
+  "rubocop -Da --force-exclusion #{files}"
```
orenyk: Uncommit?

orenyk:
Also unblocked now so we can add the intro text in to the student dashboard as well as the floor plan URL. I wonder if we should also display the student dashboard view to reps in addition to the admin one?

app/controllers/dashboards_controller.rb
esoterik: this is more concise i think

app/controllers/dashboards_controller.rb
esoterik: this is the simplest way to do it because we have to actually drop values from the `groups` hash, which isn't particularly nice to do:
```ruby
    locked_counts = groups.each_with_object({}) do |(k, v), h|
      h[k] = v.select(&:locked?)
    end.transform_values(&:count)
```

app/views/dashboards/show.html.erb
esoterik: no, i want them both to show for reps so i think this is the easiest way?

app/views/draws/_group_report.html.erb
esoterik: yup, that's why this is blocked i think

lib/tasks/style_checker.rake
esoterik: hahaha yeah

app/controllers/dashboards_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+    groups.default = []
+    group_counts = groups.transform_values(&:count)
+    group_counts.default = 0
+    sizes = (suite_counts.keys + group_counts.keys).uniq.sort
+    diff = sizes.map { |s| [s, suite_counts[s] - group_counts[s]] }.to_h
+
+    # locked_counts = draw.groups.where(status: 'locked').group(:size).count
+    locked_counts = groups.each_with_object({}) do |(k, v), h|
+      h[k] = v.select(&:locked?).size
+    end
```
esoterik: @orenyk which do we like better? your call

esoterik:
@orenyk ready for rereview!

app/controllers/dashboards_controller.rb
orenyk: `groups.transform_values { |v| v.select(&:locked?).size }`?

app/views/dashboards/show.html.erb
orenyk: Yea ignore me, I totally misread this at 2:41 AM 😛.

orenyk: Let's switch the order maybe?

spec/features/dashboards/dashboards_spec.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.feature 'Dashboards' do
+  context 'admin' do
+    it 'renders' do
+      create_draws
+      log_in FactoryGirl.create(:admin)
+      expect(page).to have_content('Vesta')
```
orenyk: So I assume the plan is to either expand these or rework the whole thing when we release for real?

app/views/dashboards/show.html.erb
```diff
@@ -0,0 +1,44 @@
+<h1>Vesta</h1>
+<% unless current_user.student? %>
+  <% if @draws.empty? %>
+    <p>Welcome to Vesta! Use the navbar above to manage housing inventory and create draws!</p>
+  <% else %>
+    <% @draw_hash.each do |draw, hash| %>
+      <div class="card">
+        <div class="card-divider">
+          <h4><%= link_to draw.name, draw_path(draw) %> (<%= draw.status %>)</h4>
```
orenyk: We should call `humanize` on the status.

app/views/dashboards/show.html.erb
```diff
@@ -0,0 +1,44 @@
...
+        </div>
+        <div class="card-section">
+          <%= render hash[:partial], draw: draw, hash: hash %>
+        </div>
+      </div>
+    <% end %>
+  <% end %>
+<% end %>
+<% unless current_user.admin? %>
+  <div>
```
orenyk: This should definitely be wrapped in a card for consistency with the heading "Welcome to Vesta" (as discussed).

app/views/dashboards/show.html.erb
```diff
@@ -0,0 +1,44 @@
...
+        <div class="card-section">
+          <% if @room.present? %>
+            <p>Your room is <%= @room.number_with_type %></p>
+          <% else %>
+            <p>Your leader has not yet submitted room assignments!</p>
+          <% end %>
+        </div>
+      </div>
+    <% end %>
+    <%= render 'groups/group_details', group: @group %>
```
orenyk: This can be removed since we have a `My Group` link in the nav - alternatively we can just add a `Your Group:` item with a link to the group page as well.

PR #481, opened 2017-03-28
Ensure group disbanding clears suite assignment

### LINKED ISSUE #478: [Vesta/production] ActionView::Template::Error: undefined method `name' for nil:NilClass
## Backtrace

    line 12 of [PROJECT_ROOT]/app/views/suites/show.html.erb: _app_views_suites_show_html_erb___1248266953607598957_47119283509660

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/50268/faults/33105941)

orenyk:
Resolves #478

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -15,7 +15,7 @@
 class Group < ApplicationRecord # rubocop:disable ClassLength
   belongs_to :leader, inverse_of: :led_group, class_name: 'User'
   belongs_to :draw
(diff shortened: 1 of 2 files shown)


orenyk:
Oh wow, I'm not sure. Good catch. Let's see if we can replicate it in PROD?

orenyk:
@esoterik back to you!

PR #503, opened 2017-03-29
[501] Prevent students from selecting suites

### LINKED ISSUE #501: Students can select suites even if the draw is not in suite selection
Oops

esoterik:
Resolves #501


orenyk:
Rubocop failures are due to errors, not actual lint issues, merging after PR build.

PR #517, opened 2017-03-29
[486] Allow lottery number removal

### LINKED ISSUE #486: Fix lottery number removal
Right now if you remove a lottery number, it gets replaced with 0 (possibly nil) which is bad because it puts that group at the start of selection

orenyk:
Resolves #486
- Fix broken test in suite split form

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 All notable changes to this project will be documented in this file.
 
 ## Unreleased
+### Fixed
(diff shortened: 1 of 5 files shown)


spec/forms/suite_split_form_spec.rb
```diff
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
```
orenyk: Remove all changes to this file.

PR #552, opened 2017-04-03
[539] Add intent counts to the intent report

### LINKED ISSUE #539: Add counts to the intent report (and maybe separate tables?)
The intent report should include the counts of how many students are in each intent type. We can also consider separating it into three tables by status.

orenyk:
Resolves #539

### DIFF
--- a/app/views/draws/intent_report.html.erb
+++ b/app/views/draws/intent_report.html.erb
@@ -12,27 +12,34 @@
     <% end %>
   </div>
   <hr />
-  <table>
-    <thead>
-      <tr>
-        <td>Last Name</td>
-        <td>First Name</td>
-        <td>Intent</td>
-      </tr>
-    </thead>
-    <tbody>
-      <% @students.each do |student| %>
-        <tr class="<%= student.intent %>">
-          <td data-role="student-last_name"><%= student.last_name %></td>
-          <td data-role="student-first_name"><%= student.first_name %></td>
-          <% if policy(student).update_intent? %>
-            <td class="intent-form" id="intent-form-<%= student.id %>" data-role="student-intent"><%= render 'intent_form', user: student %></td>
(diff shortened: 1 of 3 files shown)


app/views/draws/intent_report.html.erb
```diff
@@ -12,27 +12,34 @@
...
-          <td data-role="student-first_name"><%= student.first_name %></td>
-          <% if policy(student).update_intent? %>
-            <td class="intent-form" id="intent-form-<%= student.id %>" data-role="student-intent"><%= render 'intent_form', user: student %></td>
-          <% else %>
-            <td data-role="student-intent"><%= student.intent.humanize %></td>
+  <% if @students_by_intent.empty? %>
+    <p>There are no students in this draw yet!</p>
+  <% else %>
+    <% @students_by_intent.keys.sort.reverse.each do |intent| %>
+      <h3><%= intent %> (<%= pluralize(@students_by_intent[intent].size, 'student') %>)</h3>
```
esoterik: hmm, can we use the `IntentMetricsQuery`?

orenyk: We could, but that would be a separate query and we're already collecting the actual arrays. Do you feel strongly about it?

esoterik: this just feels like too much logic in the view and we've already built out and tested this exact behavior elsewhere

orenyk: I started implementing this with `IntentMetricsQuery`, then realized that it causes issues with the filter actions since we're going to want to use the `@students_by_intent` keys for iterating instead of _all_ the intents in the draw. Instead, I just created an intent metrics hash in the controller by transforming the values of the `@students_by_intent` hash and adding some `order` calls. Update and pushed.

PR #564, opened 2017-04-04
[561] Add building names to suite labels where appropriate

### LINKED ISSUE #561: Include the building name when displaying the suite number
This can either be an additional method (e.g. `number_with_building`) or we can overwrite the accessor - either way, it's necessary since we can have two suites with the same number that we will want to distinguish during selection and draw assignment.

orenyk:
Resolves #561


esoterik:
also, changelog?

PR #589, opened 2017-04-14
[585] Allow admins to do group CRUD as long as the draw is not a draft

### LINKED ISSUE #585: Allow admins to always do all group actions
This came up during the stiles draw where people switched groups during lottery selection. 

Reps shouldn't have this ability, but admins should be able to disband, edit, and create groups always.

esoterik:
Resolves #585
  - Removes DrawPolicy#create_new_group?

### DIFF
--- a/app/services/updaters/group_updater.rb
+++ b/app/services/updaters/group_updater.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
-#
+
 # Service object to update groups
 class GroupUpdater
   # Allows calling of :update on new instance of the class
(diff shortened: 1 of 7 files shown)


app/services/updaters/group_updater.rb
```diff
@@ -39,6 +39,7 @@ def process_params(params)
     @pending_users = { add: users(:member_ids), remove: users(:remove_ids) }
     delete_user_id_keys_from_params
     delete_leader_id_if_empty
+    delete_size_if_empty
```
orenyk: I think I'd prefer to simply hide the field on the form if the group (or its size) is locked - that seems like it makes more sense to me as opposed to passing empty params.

spec/policies/draw_policy_spec.rb
```diff
@@ -107,7 +107,7 @@
     permissions :index? do
       it { is_expected.to permit(user, Draw) }
     end
-    permissions :group_actions?, :create_new_group? do
```
orenyk: Did we not test for this for students?

PR #593, opened 2017-04-14
[590] Hide College Info on Home if no colleges

### LINKED ISSUE #590: "Too many redirects" on home page
Can be reproduced by
1. sign out
2. go to homepage

Expected: see homepage, but logged out
Actual: Chrome error saying it redirects too many times

the homepage is throwing a 302 error which is then redirecting back to the homepage. I suspect it is caused by somehow not having permissions to access the home page while logged out?

Workaround:
1. go to /users/sign_in
2. sign in

AlanLiu96:
Resolves #590

Hides colleges if no college associated
  - fixes a bug with redirects/unauthorized

### DIFF
--- a/app/views/pages/home.html.erb
+++ b/app/views/pages/home.html.erb
@@ -2,7 +2,7 @@
   <h1>Welcome to Vesta</h1>
   <p>
   Vesta is a web app that manages the housing process for Yale's residential colleges.
-  <% unless @current_college.name.empty? || @current_college.admin_email.empty? %>
+  <% unless !@current_college || @current_college.name.empty? || @current_college.admin_email.empty? %>
     This site was set up for <b><%= @current_college.name %></b> and you can log in by clicking <%= link_to 'here', new_user_session_path %>. If you don't have access and believe that you should, please contact the <%= link_to 'site administrators', "mailto:#{@current_college.admin_email}" %>.
   <% end %>
   </p>


app/views/pages/home.html.erb
```diff
@@ -2,7 +2,7 @@
   <h1>Welcome to Vesta</h1>
   <p>
   Vesta is a web app that manages the housing process for Yale's residential colleges.
-  <% unless @current_college.name.empty? || @current_college.admin_email.empty? %>
+  <% unless !@current_college || @current_college.name.empty? || @current_college.admin_email.empty? %>
```
orenyk: This shouldn't change anything, the controller action is going to run regardless because it's a `before_action` in the controller.

orenyk (changes requested):
I'm not sure I agree with the diagnosed cause of this issue - the view renders fine when there is no authenticated user to begin with, so there must be something specific to the signout process that causes it.

FWIW, I was unable to replicate this on the live pilot instances - I'm thinking it must have something to do with the redirect following logout on Devise since currently CAS logout ends up taking you to the Yale CAS logout page and not redirecting back to Vesta.

Hope that helps!

AlanLiu96:
Hmmm, after testing it some more, I think the issue was that the colleges table wasn't populated causing the before action to fail. Is this something worth testing/addressing? It's in the seed script now, but wasn't in the migrations.

PR #594, opened 2017-04-15
[257] Add suite unmerging

### LINKED ISSUE #257: Store Original Suite After Merging
We should store the original suite after merging suites together to implement restoring all original suites of a merged suite.

esoterik:
Resolves #257
  - Adds SuiteUnmerger
  - Adds original suite attribute to rooms
  - Adds Room#store_original_suite!

### DIFF
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -7,6 +7,8 @@
 # @attr [Integer] beds The number of beds within the room.
 #   For common rooms, this is 0.
 # @attr [Suite] suite belongs_to association for the Suite that the room is in.
+# @attr [String] original_suite The original suite that the room was in. Is an
+#   empty string unless the room belongs to a merged suite.
 class Room < ApplicationRecord
   belongs_to :suite
   has_many :users, dependent: :nullify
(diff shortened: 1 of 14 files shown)


spec/forms/suite_merger_form_spec.rb
```diff
@@ -89,6 +92,15 @@
         result = described_class.submit(suite: suite, params: params)
         expect(result[:object].draws.map(&:id)).to include(draw.id)
       end
+      it 'calls :store_original_suite! on the rooms' do
```
esoterik: @orenyk  I have no idea why this is failing.

murphy-ariel:
What would happen if a suite is merged, unmerged, merged again, then unmerged again? In app/models/rooms it seems like the suite would be left with outdated info in the :original_suite attribute.

app/models/room.rb
```diff
@@ -40,6 +42,16 @@ def number_with_type
     "#{number} (#{type})"
   end
 
+  # Sets the original suite to the current suite, unless it is already set
+  #
+  # @params [Hash] attrs Other attributes to update
+  #
+  # @return [Boolean] Whether or not the update was successful
+  def store_original_suite!(**attrs)
```
murphy-ariel: What would happen if a suite is merged, unmerged, merged again, then unmerged again? It seems like the suite would be left with outdated info in the :original_suite attribute.

esoterik: We want the original suite to always be the first suite the rooms belongs into, so once this attribute is set it shouldn't change.

In that scenario, `original_suite` would always be the room's *first* suite, not any of the ones that it belonged to during previous merges.

